PR #9, opened 2016-01-07
Changes to support specify IP addresses for the container

aboch:
Needed by docker/docker#19001

### DIFF
--- a/client/container_create.go
+++ b/client/container_create.go
@@ -7,16 +7,18 @@ import (
 
 	"github.com/docker/engine-api/types"
 	"github.com/docker/engine-api/types/container"
+	"github.com/docker/engine-api/types/network"
 )
 
 type configWrapper struct {
 	*container.Config
-	HostConfig *container.HostConfig
+	HostConfig       *container.HostConfig
+	NetworkingConfig *network.NetworkingConfig
 }
 
 // ContainerCreate creates a new container based in the given configuration.
 // It can be associated with a name, but it's not mandatory.
-func (cli *Client) ContainerCreate(config *container.Config, hostConfig *container.HostConfig, containerName string) (types.ContainerCreateResponse, error) {
+func (cli *Client) ContainerCreate(config *container.Config, hostConfig *container.HostConfig, networkingConfig *network.NetworkingConfig, containerName string) (types.ContainerCreateResponse, error) {
 	var response types.ContainerCreateResponse
 	query := url.Values{}
 	if containerName != "" {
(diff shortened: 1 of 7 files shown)


calavera:
LGTM

client/container_create.go
```diff
@@ -7,16 +7,18 @@ import (
...
 	*container.Config
-	HostConfig *container.HostConfig
+	HostConfig       *container.HostConfig
+	NetworkingConfig *network.NetworkingConfig
 }
 
 // ContainerCreate creates a new container based in the given configuration.
 // It can be associated with a name, but it's not mandatory.
-func (cli *Client) ContainerCreate(config *container.Config, hostConfig *container.HostConfig, containerName string) (types.ContainerCreateResponse, error) {
+func (cli *Client) ContainerCreate(config *container.Config, hostConfig *container.HostConfig, networkingConfig *network.NetworkingConfig, containerName string) (types.ContainerCreateResponse, error) {
```
calavera: @aboch I think this is not going to implement the interface that we just added in https://github.com/docker/engine-api/blob/master/client/interface.go

Do you mind changing the interface?

jfrazelle:
LGTM once that is updated

types/network/network.go
```diff
@@ -31,3 +40,26 @@ type EndpointSettings struct {
...
+}
+
+// NetworkingConfig represents the container's networking configuration for each of its interfaces
+// Carries the networink configs specified in the `docker run` and `docker network connect` commands
+type NetworkingConfig struct {
+	EndpointsConfig map[string]*EndpointSettings // Endpoint configs for each conencting network
+}
+
+// HasUserDefinedIPAddress returns whether the passed endpoint configuration contains IP address configuration
+func HasUserDefinedIPAddress(epConfig *EndpointSettings) bool {
```
calavera: This can easily be implemented in docker.

types/network/network.go
```diff
@@ -31,3 +40,26 @@ type EndpointSettings struct {
 	GlobalIPv6PrefixLen int
 	MacAddress          string
 }
+
+// CleanOperationalData resets the operational data for this endpoint
+func (es *EndpointSettings) CleanOperationalData() {
```
calavera: do we need this function here? I'd rather not add methods that are not part of the api. This can easily be implemented in docker.

types/network/network.go
aboch: Ok, I will move it to docker

types/network/network.go
aboch: Ok

client/container_create.go
aboch: Ok

calavera:
:green_heart: :exclamation:

PR #10, opened 2016-01-07
Add authentication to the docker client library

nalind:
These are the client API library changes from https://github.com/docker/docker/pull/18514 which would now be made here: the client learns to use cookies, to use supplied callbacks to obtain user names, passwords, and tokens to use for Basic and Bearer authentication, and to attempt Negotiate authentication if it's been enabled at run-time.

### DIFF
--- a/client/authn/authn.go
+++ b/client/authn/authn.go
@@ -1,0 +1,63 @@
+package authn
+
+import (
+	"net/http"
+	"strings"
+)
+
+// Logger is an interface for debug and logging callbacks.
+type Logger interface {
+	Debug(formatted string)
+	Info(formatted string)
+	Error(formatted string)
+}
+
+// ignorelogs is an implementation of Logger that ignores everything it's
+// passed.  We use it as the default set of logging callbacks if none are set.
+type ignorelogs struct {
+}
+
+func (l *ignorelogs) Debug(formatted string) {}
+func (l *ignorelogs) Info(formatted string)  {}
+func (l *ignorelogs) Error(formatted string) {}
+
+// authResponder is an interface that wraps the scheme,
+// authRespond, and authCompleted methods.
+//
+// At initialization time, an implementation of authResponder should register
+// itself by calling registerAuthResponder.
+type authResponder interface {
+	// Scheme should return the name of the authorization scheme for which
+	// the responder should be called.
+	scheme() string
+	// authRespond, given the authentication header value associated with
+	// the scheme that it implements, can decide if the request should be
+	// retried.  If it returns true, then the request is retransmitted to
+	// the server, presumably because it has added an authentication header
+	// which it believes the server will accept.
+	authRespond(challenge string, req *http.Request) (bool, error)
+	// AuthCompleted, given a (possibly empty) WWW-Authenticate header and
+	// a successful response, should decide if the server's reply should be
+	// accepted.
+	authCompleted(challenge string, resp *http.Response) (bool, error)
+}
+
+// authResponderCreators functions create a new authResponder.
+var authResponderCreators = []func(logger Logger, authers []interface{}) authResponder{}
+
+// Run through all of the registered responder creators and build a map of
+// names-to-responder-instances.
+func createAuthResponders(logger Logger, authers []interface{}) map[string]authResponder {
+	if logger == nil {
+		logger = &ignorelogs{}
+	}
+	ars := make(map[string]authResponder)
+	for _, arc := range authResponderCreators {
+		responder := arc(logger, authers)
+		if responder != nil {
+			scheme := strings.ToLower(responder.scheme())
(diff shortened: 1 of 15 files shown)


client/authn.go
```diff
@@ -0,0 +1,199 @@
...
+		if cookies := resp.Cookies(); cookies != nil {
+			cli.jar.SetCookies(req.URL, cookies)
+		}
+	}
+	return resp, err
+}
+
+// doWithAuthn calls Do(), and handles any "unauthorized" errors
+// which it returns by retrying it with authentication.
+func (cli *Client) doWithAuthn(doer doer, req *http.Request, body []byte) (*http.Response, error) {
```
calavera: this should take a reader as a body, not `[]byte`.

calavera:
I think we should avoid adding logging unless we can provide a way to change it. I like logrus, but I don't want to force everyone to use it.

nalind:
Makes sense.  The log messages are very useful for troubleshooting things, so I'll tweak it to take optional logging callbacks.

client/authn.go
nalind: Hmm, Reader isn't going to be good enough, since if we get a 401 error and have to reattempt a request with authentication, we'll need to be able to supply the same POST data that we sent the first time.
We'd been reading it all into a byte array, first thing, and just referring to it every time, but that can get expensive.
I think we can add a rewindable wrapper around the body Reader, one that hangs on to everything that's read through it, so that it can be explicitly rewound before every attempt at hitting the server.  If the server comes back with an error fast enough, maybe the client won't have run through too much of the input data.

calavera:
I'd rather have a single method to add a logger. We should define an interface that a logger must implement, like this:

```go
type Logger interface {
   Debugf(...)
   Infof(...)
   Errorf(...)
}

type APIClient interface {
   SetLogger(Logger)
}
```
   
Can we put all this new auth logic into a package? something like `client/auth`?

I also have mixed feelings about the implementation, sorry :pensive: 
I wonder if we can design all this like client middlewares to minimize the logic sprawl into the client?

We could design a `CookieMiddleware` and an `AuthMiddleware` such as the cookie middleware injected cookies to the request, and the auth middleware handle the responses. And them add them to the client with something like `client.AddMiddleware(CookieMiddleware, AuthMiddleware)`. The client would execute the middlewares in a chain and would inject itself as last executor. That way it would not need to know anything about cookie jars or auth challenges.

I basically don't want to add these 4 functions to the interface, to be clear :laughing: 

```go
SetBasicAuth(get func(challenge string) (string, string, error))
SetBearerAuth(get func(challenge string) (string, error))
SetCookieJar(jar http.CookieJar)
SetNegotiateAuth(enabled bool)
```

I only want to add 1: `SetMiddleware(m ...Middleware)` or `SetAuth(m ...Auth)` or something similar.

dnephin:
I like the idea of middlewares. With them I think the authentication could be in a separate package and only the interface would need to be in engine-api ?

nalind:
Okay, reworked a lot of the logic to wrap the Do() method used for communicating with the server.  Logging callbacks are used if a Logger interface is passed in.  The authentication bits call out to a specified provider of an appropriate interface when it needs specific information.

calavera:
I like this A LOT, thanks @nalind.

I'm going to hold on merging it because code freeze for docker 1.10 is tomorrow and I'd like to make sure we don't find inconsistencies between versions.

cyphar:
I don't understand what we get from this that we don't get from AuthZ?

nalind:
The authorization plugins currently have no knowledge of who issued the request that they're being asked to authorize, so they're not able to base any part of their decisions on that.  The daemon-side plumbing in https://github.com/docker/docker/pull/18514 is attempting to supply that information.

In some cases, the logic that we add to the daemon in that PR can accomplish that without client help.  It can dig identity information out of Unix sockets (when the client connects over a Unix socket) and by examining client certificates with the help of plugins (when client certificates are being used).  It can't authenticate clients using passwords or Kerberos, though, without explicitly asking the client to authenticate.

When the daemon does ask the client to authenticate, it's done in ways that are defined for HTTP.  This PR is an attempt to provide consumers of the client library with a way to supply necessary information to the library, so that the library can successfully authenticate when the daemon asks it to.

cyphar:
@nalind Ah okay, cool. :D

calavera:
I'm going to carry this today-ish. There have been some changes in the client that affect this. I'll close this PR when I have the new one open.

nalind:
Should I take a stab at the current conflict, then?  I didn't see the mockable transport interface changing much in the hijack code paths, so not sure what we're planning to do there.

calavera:
@nalind I'm fixing the conflicts, I'll ping you when it's done

calavera:
Please, refer to #72 to follow up.

PR #11, opened 2016-01-07
Add a specific config for the update command

vdemeester:
This is related to docker/docker#19104.

> This is a proposal for #18957, to reduce what we send to the update endpoint. This also make it clear what is mutable in the container, and untie it from Config/HostConfig. It is still easy to add more attributes
>
> This allows use to control more what is or is not mutable in a container and remove the use of the internal HostConfig struct to be used 🐳.

/cc @calavera @tiborvass

### DIFF
--- a/client/container_update.go
+++ b/client/container_update.go
@@ -5,8 +5,8 @@ import (
 )
 
 // ContainerUpdate updates resources of a container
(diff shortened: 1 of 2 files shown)


tiborvass:
LGTM

calavera:
LGTM

PR #18, opened 2016-01-08
Send only the endpoint settings to connect to a network.

calavera:

### DIFF
--- a/client/interface.go
+++ b/client/interface.go
@@ -58,7 +58,7 @@ type APIClient interface {
 	ImageSave(imageIDs []string) (io.ReadCloser, error)
 	ImageTag(options types.ImageTagOptions) error
 	Info() (types.Info, error)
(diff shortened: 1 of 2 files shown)


tiborvass:
LGTM

client/interface.go
```diff
@@ -58,7 +58,7 @@ type APIClient interface {
 	ImageSave(imageIDs []string) (io.ReadCloser, error)
 	ImageTag(options types.ImageTagOptions) error
 	Info() (types.Info, error)
-	NetworkConnect(networkID, containerID string, config *network.NetworkingConfig) error
+	NetworkConnect(networkID, containerID string, config *network.EndpointSettings) error
```
aboch: If we are never going to support connecting to multiple networks on a single call, it is fine to remove this.

PR #21, opened 2016-01-08
Change OomKillDisable to be pointer

hqhq:
It's like `MemorySwappiness`, the default value has specific
meaning (default false means enable oom kill).

We need to change it to pointer so we can update it after
container is created.

Part of: https://github.com/docker/docker/pull/19014

Signed-off-by: Qiang Huang <h.huangqiang@huawei.com>


cpuguy83:
LGTM

calavera:
LGTM

PR #28, opened 2016-01-09
Added missing omitempty JSON tag

### LINKED ISSUE #25: does it make sense to omitempty json attributes?
I was just wondering it was possible to have some consistence since some attributes have "omitempty" and others doesn't. For example: https://github.com/docker/engine-api/blob/master/types/types.go#L296
Are there any cons in using the "omitempty" tag?

aanm:
Fixes #25 

Signed-off-by: André Martins <aanm90@gmail.com>

### DIFF
--- a/types/configs.go
+++ b/types/configs.go
@@ -11,44 +11,44 @@ import (
 
 // ContainerCreateConfig is the parameter set to ContainerCreate()
 type ContainerCreateConfig struct {
-	Name             string
-	Config           *container.Config
-	HostConfig       *container.HostConfig
-	NetworkingConfig *network.NetworkingConfig
-	AdjustCPUShares  bool
+	Name             string                    `json:",omitempty"`
+	Config           *container.Config         `json:",omitempty"`
+	HostConfig       *container.HostConfig     `json:",omitempty"`
+	NetworkingConfig *network.NetworkingConfig `json:",omitempty"`
+	AdjustCPUShares  bool                      `json:",omitempty"`
 }
 
 // ContainerRmConfig holds arguments for the container remove
 // operation. This struct is used to tell the backend what operations
 // to perform.
 type ContainerRmConfig struct {
-	ForceRemove, RemoveVolume, RemoveLink bool
+	ForceRemove, RemoveVolume, RemoveLink bool `json:",omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


calavera:
I don't really know the full repercussions of this change. I'd go for it if we could validate that everything in docker keeps working as expected.

aanm:
@calavera In theory should work, other docker-go libraries do the same. But still, how can we prove that?

calavera:
@aanm vendor this patch into docker/docker, run `make test-integration-cli`. We're good if all tests pass :smile:

aanm:
@calavera So I did that but unfortunately some tests failed:
```
FAIL: docker_api_containers_test.go:51: DockerSuite.TestContainerApiGetJSONNoFieldsOmitted

docker_api_containers_test.go:78:
    c.Fatalf("Field %s is missing and it shouldn't", f)
... Error: Field Ports is missing and it shouldn't
```

I did a little more and there's some discussion about this in https://github.com/docker/docker/pull/13703

ping @duglin @runcom

runcom:
I think it's pretty clear that we did that (in the linked PR) because we needed to maintain backward compatibility and do not drop the field if it was empty (IIRC)

aanm:
@runcom Ah! I was missing that. Somehow the statement "To be fully compatible with 1.18, SizeRw and SizeRootFs need to be omit empty." confused me.

runcom:
I'm not confident about adding all those omitempty(s), we had hassles in docker everytime we added/removed one :) not sure we should go with this PR, ping @calavera

aanm:
@runcom I agree and understand your POV. But still I think it's something that needs to be addressed and the sooner the better. I just can't understand why "default" attributes should be sent. If the tests pass, there's not need to be worried :-)

jfrazelle:
Tests don't test everything I'm pretty confident this would break things

cyphar:
The new content-addressibility stuff is particularly important in this regard: we base the hashes on the metadata -- changing what fields are omitted will cause all of that to change.

calavera:
I'm closing this PR given that a few maintainers have expressed their concerns about it. I think we should play save and don't add this if we don't really need it.

Thank you all for your comments.

PR #29, opened 2016-01-09
an option to disconnect an endpoint from a network forcefully

mavenugo:
There are cases as in https://github.com/docker/swarm/pull/1578 where swarm needs a way to
forcefully disconnect an endpoint from a network when a swarm node goes down ungracefully.

Signed-off-by: Madhu Venugopal <madhu@docker.com>

### DIFF
--- a/client/interface.go
+++ b/client/interface.go
@@ -60,7 +60,7 @@ type APIClient interface {
 	Info() (types.Info, error)
 	NetworkConnect(networkID, containerID string, config *network.EndpointSettings) error
 	NetworkCreate(options types.NetworkCreate) (types.NetworkCreateResponse, error)
-	NetworkDisconnect(networkID, containerID string) error
+	NetworkDisconnect(networkID, containerID string, force bool) error
 	NetworkInspect(networkID string) (types.NetworkResource, error)
 	NetworkList(options types.NetworkListOptions) ([]types.NetworkResource, error)
 	NetworkRemove(networkID string) error
--- a/client/network.go
+++ b/client/network.go
@@ -42,9 +42,9 @@ func (cli *Client) NetworkConnect(networkID, containerID string, config *network
 }
 
 // NetworkDisconnect disconnects a container from an existent network in the docker host.
-func (cli *Client) NetworkDisconnect(networkID, containerID string) error {
-	nc := types.NetworkConnect{Container: containerID}
-	resp, err := cli.post("/networks/"+networkID+"/disconnect", nil, nc, nil)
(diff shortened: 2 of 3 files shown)


mavenugo:
ping @calavera @vieux

icecrime:
The `Force` parameter seems to make sense. I don't quite understand how this relates to the Swarm PR but this is out of the scope of this repo so LGTM.

mavenugo:
@icecrime docker/swarm#1578 handles the case of container rescheduling on node failures. Node Failures typically are ungraceful and hence some of the distributed container states (mainly service-name and ip-address managed by distributed IPAM) in KVStore gets locked up till the failed node comes back up again. But Swarm must reschedule the containers in the failed node in another host. But since the name and ip-address are locked up in the KVStore, swarm will not be able to effectively perform a proper re-scheduling.
The proposal is to provide an ability to cleanup the endpoint distributed states from any node in the swam cluster (not necessarily from the failed host alone).

This PR (and its corresponding changes in docker and libnetwork) will help to get there.

icecrime:
@mavenugo So if I understand correctly, this request to force clean will be sent to _any_ random node, not the one which initially created the container (as it's now unreachable).

I find this very confusing. It seems were only adding this to the API in order to easily reuse some libnetwork functions that isn't even tied to the state of the executing daemon. Basically: we're doing this because it's an easy way to call libnetwork.

It would be too hard to do differently in the current design, but it seems like a strong argument for a Swarm manager to be capable of managing such resources by itself.

mavenugo:
@icecrime your assessment is correct with respect to swarm's use case. But the same problem exists for docker engine's multi host networking as well.
We have a few issues opened due to the stake endpoint which is a result of the same case. Hence this will be an engine feature which swarm can make use of.

icecrime:
@mavenugo Can you clarify how does the problem exist for Engine? If a daemon crashes and restart, isn't it cable of doing a network cleanup without this flag?

mavenugo:
> If a daemon crashes and restart, isn't it cable of doing a network cleanup without this flag?

@icecrime Yes. But the case that we are trying to address is the case when the daemon crashes and never restarts again. The stale entries consumed by the containers are locked in place.
BTW, we didn't implement a TTL based expiry mechanism. Any TTL based mechanism on KV Store will kill performance. Especially having TTL for every container in store will bring it to a crawl.

By providing this mechanism, we are letting the engine user to take control in cases where a failed engine never starts again (maybe a bad hardware ?) :( . The same problem exists for Swarm and hence the solution makes sense for both engine and swarm.

icecrime:
@mavenugo I see, thanks.

calavera:
LGTM

PR #31, opened 2016-01-11
Change missing image detection for create

tonistiigi:
Daemon may not respond with the same image name
any more but it may be in a normalized form.
Generally checking for a image name doesn’t 
seem like a good idea because small image names
may match any possible error.

I'd be happy to only leave in `404` check if someone
is confident that this would not break somthing else.

Signed-off-by: Tonis Tiigi <tonistiigi@gmail.com>


calavera:
LGTM

PR #32, opened 2016-01-11
Revert "Send graph storage options as part of the host configuration."

tiborvass:
This reverts commit 196e724016a2bf83f5c4196a78a6689579906f93.

Reverting this because the usecase is too specific for adding
--storage-opt on container creation. We can revisit this and other
solutions after Docker's 1.10 release.

Signed-off-by: Tibor Vass <tibor@docker.com>


icecrime:
LGTM

calavera:
LGTM

PR #35, opened 2016-01-12
Make streaming operations cancellable.

### LINKED ISSUE #4: Cancellation on streaming requests can lead to connection leak
When doing a streaming request we can end-up leaking connections to the docker daemon if there's no activity on the stream. A simple example of this is calling Client.ContainerLogs() with Follow:true and stopping. No new logs must be generated from that point on. While we do have access to http.Response.Body to close it, the connection will only close on the next write error.

This [module](https://godoc.org/golang.org/x/net/context/ctxhttp) would make [cancellation](https://blog.golang.org/context) fairly easy to integrate with the current code.

calavera:
This change adds a new context argument to all streaming operations to make them cancellable.

Fixes #4.


client/image_build.go
```diff
@@ -32,7 +34,7 @@ func (cli *Client) ImageBuild(options types.ImageBuildOptions) (types.ImageBuild
 	headers.Add("X-Registry-Config", base64.URLEncoding.EncodeToString(buf))
 	headers.Set("Content-Type", "application/tar")
 
-	serverResp, err := cli.postRaw("/build", query, options.Context, headers)
+	serverResp, err := cli.postRaw(context.Background(), "/build", query, options.Context, headers)
```
dongluochen: For my education, why not pass current context?

calavera: yeah, I think this function should be allowed to pass a context too.

calavera:
I think we should have several examples on how to use the contexts before merging this in. @mathpl, I really appreciate some help with that if you're up to contribute.

mathpl:
I'd be happy to contribute. Where would you want these, doc.go or actual code under an example directory?

calavera:
@mathpl either sounds good to me. I think the advantage of having them in a `doc.go` or `_test.go` as `func Example` would be that we can show them in godoc.org.

calavera:
I added an example of canceling logs stream. I think this is ready to merge, we can add more examples later.

client/transport/cancellable/cancellable.go
```diff
@@ -0,0 +1,107 @@
...
+	}()
+
+	var resp *http.Response
+
+	select {
+	case <-ctx.Done():
+		testHookContextDoneBeforeHeaders()
+		cancel()
+		// Clean up after the goroutine calling client.Do:
+		go func() {
```
aaronlehmann: What's the motivation for spawning a new goroutine here? I would expect `Do` not to return until the cancellation is acknowledged and the response body is closed.

client/transport/cancellable/cancellable.go
```diff
@@ -0,0 +1,107 @@
...
+			if r := <-result; r.resp != nil {
+				testHookDidBodyClose()
+				r.resp.Body.Close()
```
aaronlehmann: I'm being pedantic, but I think before calling `r.resp.Body.Close()`, you need to check that either `err` is nil, or `r.resp.Body` is non-nil:

    When err is nil, resp always contains a non-nil resp.Body.

Having a non-nil `resp` does not seem to be a sufficient condition.

client/transport/cancellable/cancellable.go
calavera: https://github.com/golang/net/commit/f315505cf3349909cdf013ea56690da34e96a451

client/transport/cancellable/cancellable.go
```diff
@@ -0,0 +1,107 @@
+// Copyright 2015 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// Package cancellable provides helper function to cancel http requests.
+package cancellable
```
aaronlehmann: Looks like this is `golang.org/x/net/context/ctxhttp`. Why is it being forked instead of vendored?

If there are modifications from upstream, those should be noted.

client/transport/cancellable/cancellable.go
calavera: agreed.

client/transport/cancellable/cancellable.go
calavera: I added a not in the function declaration.

aaronlehmann:
LGTM

client/interface.go
```diff
@@ -22,40 +24,40 @@ type APIClient interface {
...
+	ContainerStats(ctx context.Context, containerID string, stream bool) (io.ReadCloser, error)
 	ContainerStart(containerID string) error
 	ContainerStop(containerID string, timeout int) error
 	ContainerTop(containerID string, arguments []string) (types.ContainerProcessList, error)
 	ContainerUnpause(containerID string) error
 	ContainerUpdate(containerID string, updateConfig container.UpdateConfig) error
 	ContainerWait(containerID string) (int, error)
-	CopyFromContainer(containerID, srcPath string) (io.ReadCloser, types.ContainerPathStat, error)
+	CopyFromContainer(ctx context.Context, containerID, srcPath string) (io.ReadCloser, types.ContainerPathStat, error)
 	CopyToContainer(options types.CopyToContainerOptions) error
```
LK4D4: Is this non-cancelable?

client/interface.go
```diff
@@ -22,40 +24,40 @@ type APIClient interface {
...
 	ContainerRestart(containerID string, timeout int) error
 	ContainerStatPath(containerID, path string) (types.ContainerPathStat, error)
-	ContainerStats(containerID string, stream bool) (io.ReadCloser, error)
+	ContainerStats(ctx context.Context, containerID string, stream bool) (io.ReadCloser, error)
 	ContainerStart(containerID string) error
 	ContainerStop(containerID string, timeout int) error
 	ContainerTop(containerID string, arguments []string) (types.ContainerProcessList, error)
 	ContainerUnpause(containerID string) error
 	ContainerUpdate(containerID string, updateConfig container.UpdateConfig) error
 	ContainerWait(containerID string) (int, error)
```
LK4D4: Looks like cancellable too.

calavera:
@LK4D4 I've made those two operations cancelable too. Thanks for checking.

LK4D4:
LGTM

PR #39, opened 2016-01-13
use defaut unix socket if DOCKER_HOST is unset

lalyos:


duglin:
I'm pretty sure we already do this, but can you add a testcase? That will show for sure whether or not there's an issue.

If there is, then you may want to look to see if you can use the funcs & constants in opts/hosts.go (like DefaultUnixSocket) so that we don't have socket address hard coded in multiple places in the code.

calavera:
@duglin we do that in docker, not in this library :stuck_out_tongue: 

@lalyos unfortunately, it might slightly more complicated than this. The docker server is multiplatform, so we should not assume that the client is talking with a Linux server. I'm okay with having a default value for that case, but we should have a default for linux, in a file called `client/client_linux.go` and a default for windows, in a file called `client/client_windows.go`.

As @duglin says, we have those constants defined in docker/docker. Maybe we could port them here and remove them from the code there.

lalyos:
@duglin I was running into this with the same test code from: #40 

```
$ GOOS=linux go build -o docker-cli-linux main.go
$ docker run \
  -v /var/run/docker.sock:/var/run/docker.sock \
  -v $PWD:/work \
  -w /work \
  alpine ./docker-cli-linux

panic: runtime error: index out of range

goroutine 1 [running]:
github.com/docker/engine-api/client.NewClient(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x14, 0x0, 0x0)
        /Users/lpapp/go/src/github.com/docker/engine-api/client/client.go:71 +0x575
```

So without this patch the the following [2 lines](https://github.com/docker/engine-api/blob/master/client/client.go#L70-L71), tries to split an empty string, a refer to the 2. element:
```
		protoAddrParts = strings.SplitN(host, "://", 2)
		proto, addr    = protoAddrParts[0], protoAddrParts[1]
```

calavera:
@lalyos you only need to put the default string as a constant in the new file. Something like:

```go
const DefaultDockerHost = "unix:///var/run/docker.sock"
```

and use it in the function, with something like:

```go
host := os.Getenv("DOCKER_HOST")
if host == "" {
  host = DefaultDockerHost
}
```

go's build toolchain takes care of the rest.

lalyos:
@calavera i moved the linux specific code to `client_linux.go`, but
- there should be a `darwin` and `windows` specific version as well
- both are boot2docker/docker_machine based, which have different defaults. Which one should be used: `tcp://192.168.59.103:2376` or `192.168.99.100` ?

calavera:
The default for those two platforms should be `tcp://127.0.0.1:2375`

lalyos:
@calavera you are right i fond the [docker topts/hosts_test.go](https://github.com/docker/docker/blob/master/opts/hosts_test.go#L18) are using those values

updated

calavera:
great, thanks!

LGTM

duglin:
Would it make sense to have this import the Docker constants so that we don't need to duplicate it in this package and then keep it in-sync as (or if) Docker ever changes them?  If its not easy to import this constant from Docker it might be good to refactor Docker to more easily allow for it.

dnephin:
I think it would make more sense for this package to be the canonical source for these. docker/docker should use the constants from this package so they can be shared with clients as well.

I suspect there will be a few more things from `opts/hosts` that will move here as well.

duglin:
It depends on which package is the dog and which is the tail  :-)
But either way, as long as its only specified in one spot and the other imports it then all is good.

calavera:
> as long as its only specified in one spot and the other imports it then all is good.

I agree with this. I feel like this should be the canonical source of information for those defaults.

@lalyos I'm not sure if you're up for the challenge. I'm still okay with merging this change and following up with another PR. What do you think @dnephin and @duglin?

duglin:
separate PR for it is fine with me.

dnephin:
Not sure I understand what's to change. Is separate PR to update docker/docker to use these constants?

calavera:
> Is separate PR to update docker/docker to use these constants?

Yeah. There are a few other constants in docker that we might want to port here too. So all of them are in one place.

lalyos:
I've updated the constant name to `DefaultDockerHost` so its public now.

Regarding the separate PR: are you suggesting to open a PR in `https://github.com/docker/docker` and refer to `DefaultDockerHost` const from here, right?

lalyos:
Ok, so Travis is happy now, this can be merged, and I will open a PR for docker, and refer to this issue.

PR #46, opened 2016-01-14
Adding CustomStatus field for /info endpoint

nishanttotla:
This is for use by Docker Swarm, but generic so that other tools can use it too.

Please check related issues: https://github.com/docker/docker/issues/19277 and https://github.com/docker/swarm/issues/1625

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -198,7 +198,7 @@ type Info struct {
 	Images             int
 	Driver             string
 	DriverStatus       [][2]string
-	CustomStatus       [][2]string
+	SystemStatus       [][2]string
 	Plugins            PluginsInfo
 	MemoryLimit        bool
 	SwapLimit          bool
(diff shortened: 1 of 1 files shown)


types/types.go
```diff
@@ -198,6 +198,7 @@ type Info struct {
 	Images             int
 	Driver             string
 	DriverStatus       [][2]string
+	CustomStatus       [][2]string
```
calavera: why is it a slice of slices of two elements? it sounds like you want a map.

nishanttotla: @calavera we want the output to be displayed in the same order it was added. You're right that this is effectively a map, but using a slice of slices gives more control to the tool using `CustomStatus` wrt how it will be displayed.

calavera: it makes sense, can we call it `CurrentStatus` or `SystemStatus`? I feel like `Custom` is very broad.

nishanttotla: I'll update the PR to `SystemStatus`, but it would nice to have some comments that specify what it does. Is there a recommended location for that, because none of the other fields have attached comments.

calavera:
LGTM

dongluochen:
LGTM

PR #54, opened 2016-01-20
Allow to configure the client with a given http.Client.

### LINKED ISSUE #24: Unable to mock client in unit tests
Hey, I was trying to mock the docker client behavior by overwriting the httpclient. Unfortunately the httpclient is private https://github.com/docker/engine-api/blob/master/client/client.go#L29
Anyway to workaround this?
Thanks!

### LINKED ISSUE #48: Passing an empty host parameter to the NewClient function crashes terribly
```
panic: runtime error: index out of range

github.com/docker/engine-api/client.NewClient(0x0, 0x0, 0x42e060, 0x5, 0xc8200ca750, 0xc82000b260, 0x0, 0x0, 0x0)
	/Users/rob/workspace/go/src/github.com/docker/engine-api/client/client.go:70 +0x586
```
At least a decent error message should be given but I think that using the docker default (`unix:///var/run/docker.sock`)when the host is empty is even a better option.

calavera:
- Give more control over the http connection to the library callers.
- Validate the host provided when the client is created.

I've added a very simple test as a reference for future mocked tests.

Fixes #24
Fixes #48


dongluochen:
LGTM.

MHBauer:
Makes sense to me. I like allowing the custom transport. It would be nice to have a test for that, but I can't think of anything that's not trivially true and adds to verification.

calavera:
> Makes sense to me. I like allowing the custom transport. It would be nice to have a test for that, but I can't think of anything that's not trivially true and adds to verification.

@MHBauer in theory, this will allow us to start adding more tests :smile:

calavera:
I've added a very simple test as a reference for future mocked tests.

icecrime:
LGTM

PR #56, opened 2016-01-21
Clean unused functions

WeiZhang555:
This is a duplicate function of `docker/runconfig/hostconfig_unix.go`, and we shouldn't call it from client side because it's related to platform. 

For example, if we have docker daemon running on Linux platform and client running on Windows platform, it's unmeaning for client to call `IsPreDefinedNetwork`.

Actually I think we can remove the whole file for same reason, and also some other functions. What do you think of this? @calavera 

### DIFF
--- a/types/container/hostconfig_unix.go
+++ b/types/container/hostconfig_unix.go
@@ -72,12 +72,6 @@ func (n NetworkMode) IsUserDefined() bool {
 	return !n.IsDefault() && !n.IsBridge() && !n.IsHost() && !n.IsNone() && !n.IsContainer()
 }
 
-// IsPreDefinedNetwork indicates if a network is predefined by the daemon
-func IsPreDefinedNetwork(network string) bool {
-	n := NetworkMode(network)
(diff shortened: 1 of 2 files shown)


calavera:
I think it makes sense to remove this functions. I rather not remove the whole files for now, until we're certain this is safe. I understand your point, we've had similar discussions before, I'd like to have someone's else eyes on it first.

dnephin:
I think removing this function and `IsUserDefined()` makes sense. That logic belongs in the daemon. 

I think the rest of the file is really just implementing a nicer interface for an enum and it's safe to call it from the client (although maybe not completely necessary).

WeiZhang555:
Yes, the reason why I'm trying to remove this function is I made some modifications once in https://github.com/docker/docker/pull/19431, since this is unmeaning for client to call it, I think we'd better remove it, or we have to modify this duplicate function every time Docker makes some changes on it.

WeiZhang555:
And thank you for help review! :smile:

PR #57, opened 2016-01-21
Clean README

WeiZhang555:
Server package doesn't exist any more

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,12 +47,6 @@ func main() {
 
 The types package includes all typed structures that client and server serialize to execute operations.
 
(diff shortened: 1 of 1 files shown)


calavera:
Sorry, the server package is still pending to be moved here, as it says in:

```
This package is still pending to be extracted from the Docker engine.
```

I'm okay with removing this from the readme, we'll have to rewrite it again anyways.

WeiZhang555:
Ah, I didn't notice it, and now I know.
It's all OK, I'll close it now. :smile:

PR #58, opened 2016-01-21
Verify auth errors for InternalServerError errors.

calavera:
Stream operations might hide authentication errors in the stream.
Verify that errored operations, like errored pulls, are not
authentication errors in reality.

Patch for docker/docker#19523

/cc @aaronlehmann, @tonistiigi

### DIFF
--- a/client/request.go
+++ b/client/request.go
@@ -174,3 +174,26 @@ func isTimeout(err error) bool {
 	t, ok := e.(timeout)
 	return ok && t.Timeout()
 }
+
+// isUnauthorized returns true if an operation fails with
+// and authorization error.
+// If the server returns a 500 error, we check the error message
+// response because the real error could have been streamed there.
+func isUnauthorized(resp *serverResponse, err error) bool {
+	switch resp.statusCode {
+	case http.StatusUnauthorized:
+		return true
+	case http.StatusInternalServerError:
+		if err == nil {
+			return false
+		}
+		loError := strings.ToLower(err.Error())
+		if strings.Contains(loError, "authentication is required") ||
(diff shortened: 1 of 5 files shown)


client/request.go
```diff
@@ -174,3 +174,26 @@ func isTimeout(err error) bool {
...
+		return true
+	case http.StatusInternalServerError:
+		if err == nil {
+			return false
+		}
+		loError := strings.ToLower(err.Error())
+		if strings.Contains(loError, "authentication is required") ||
+			strings.Contains(loError, "status 401") ||
+			strings.Contains(loError, "unauthorized") ||
+			strings.Contains(loError, "status code 401") {
```
aaronlehmann: I thought the purpose of this string grepping was to handle cases where the HTTP handler had already written a 200 status, as mentioned here: https://github.com/calavera/docker/blob/e78f02c4dbc3cada909c114fef6b6643969ab912/api/client/utils.go#L151

If this isn't the case, wouldn't it be cleaner to just change the daemon to send 401 instead of 500?

calavera: > If this isn't the case, wouldn't it be cleaner to just change the daemon to send 401 instead of 500?

yes.

I think the behavior described in that commit doesn't apply anymore to this library. All the body decoding has been delegated to the docker client. I'll open a PR in docker/docker to handle that case.

calavera: I don't really know when that can happen either, the PR that introduced that logic is not very clear about it either:

https://github.com/docker/docker/pull/10050/files

calavera:
We're going to make the daemon to return the right response status. Duh.

PR #60, opened 2016-01-23
client: fix conditional compilation for freebsd

cyphar:
The DockerDefaultHost was unset on freebsd. Fix this by adding a new
client_freebsd.go with the same default as the others.

Signed-off-by: Aleksa Sarai <asarai@suse.com>

/cc @calavera @runcom @vdemeester @crosbymichael

### DIFF
--- a/client/client_darwin.go
+++ b/client/client_darwin.go
@@ -1,6 +1,0 @@
-// +build darwin
-
-package client
-
-// DefaultDockerHost defines os specific default if DOCKER_HOST is unset
-const DefaultDockerHost = "tcp://127.0.0.1:2375"
--- a/client/client_nounix.go
+++ b/client/client_nounix.go
@@ -1,6 +1,6 @@
-// +build linux
+// +build windows darwin
 
 package client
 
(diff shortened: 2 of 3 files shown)


cyphar:
I changed it to just make a new file rather than consolidating them. Makes more sense in context.

runcom:
I don't think this is right, we should have

`client_unix.go` with `unix://...` and `// +build linux freebsd`
`client_nounix.go` with `tcp://...` and `// +build windows darwin`

(or any combination of the above as you guys prefer..)

I'm saying this because before the port of docker's types here in engine-api we had https://github.com/docker/docker/blob/v1.9.1/opts/hosts_unix.go which clearly compiled `unix://...` for freebsd (if I'm not wrong reading the code)

no big deal though, since anyone seems to use freebsd :(

cyphar:
@runcom Done. :P

vdemeester:
:+1: LGTM 🐹

runcom:
@cyphar I based my thoughts on old code so maybe, as you said, unix do not make sense in freebsd w/o daemon and this was actually a fix. Ping @calavera who might know

calavera:
LGTM

PR #61, opened 2016-01-25
Remove duplicated code

thaJeztah:
This removes duplicated code

Also removed a redundant check if err is nil, because we returned an "error" or "nil", so if err is actually nil, we're good to go as well.

### DIFF
--- a/client/container_create.go
+++ b/client/container_create.go
@@ -39,18 +39,7 @@ func (cli *Client) ContainerCreate(config *container.Config, hostConfig *contain
 		return response, err
 	}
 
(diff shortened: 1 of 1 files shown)


thaJeztah:
Travis is failing, and mentions "No environment variables set";

    The command "go get -t ./..." failed and exited with 2 during .

vdemeester:
LGTM 🐯

thaJeztah:
@vdemeester PTAL; I pushed one more minor change; I removed the defer, because it's no longer needed

vdemeester:
@thaJeztah ah cool :+1:

duglin:
LGTM

PR #64, opened 2016-01-27
Add transport types to send to backends.

calavera:
We need these types for docker/docker#19772.

### DIFF
--- a/types/backend/backend.go
+++ b/types/backend/backend.go
@@ -1,0 +1,62 @@
+package backend
+
+import (
+	"io"
+	"net/http"
+	"time"
+)
+
+// ContainersConfig is the filtering specified by the user to iterate over containers.
+type ContainersConfig struct {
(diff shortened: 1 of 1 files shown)


dongluochen:
@calavera Want to confirm these are migration, no real data structure change. Is that right?

calavera:
that's correct

dongluochen:
LGTM.

vdemeester:
@calavera what is the rational of putting these in the API if it's only used on the server side ? (I might miss something :stuck_out_tongue_closed_eyes:)

MHBauer:
I'm kind of surprised this is all coming over before the server api stuff. I'll go check on https://github.com/docker/docker/pull/19772.

calavera:
> I'm kind of surprised this is all coming over before the server api stuff. I'll go check on docker/docker#19772.

I'm surprised too, tbh. Just wanted to be practice, but it might not be worth it.

PR #65, opened 2016-01-28
Add a quiet flag to image load for adding progress bar to load

coolljt0725:
Signed-off-by: Lei Jitang <leijitang@huawei.com>

related to https://github.com/docker/docker/pull/17329

### DIFF
--- a/client/image_load.go
+++ b/client/image_load.go
@@ -12,8 +12,13 @@ import (
 // ImageLoad loads an image in the docker host from the client host.
 // It's up to the caller to close the io.ReadCloser returned by
 // this function.
-func (cli *Client) ImageLoad(ctx context.Context, input io.Reader) (types.ImageLoadResponse, error) {
-	resp, err := cli.postRaw(ctx, "/images/load", url.Values{}, input, nil)
+func (cli *Client) ImageLoad(ctx context.Context, input io.Reader, quiet bool) (types.ImageLoadResponse, error) {
+	v := url.Values{}
+	v.Set("quiet", "0")
+	if quiet {
+		v.Set("quiet", "1")
+	}
+	resp, err := cli.postRaw(ctx, "/images/load", v, input, nil)
 	if err != nil {
 		return types.ImageLoadResponse{}, err
 	}
--- a/client/interface.go
+++ b/client/interface.go
@@ -52,7 +52,7 @@ type APIClient interface {
 	ImageImport(ctx context.Context, options types.ImageImportOptions) (io.ReadCloser, error)
 	ImageInspectWithRaw(imageID string, getSize bool) (types.ImageInspect, []byte, error)
 	ImageList(options types.ImageListOptions) ([]types.Image, error)
-	ImageLoad(ctx context.Context, input io.Reader) (types.ImageLoadResponse, error)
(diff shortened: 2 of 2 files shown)


vdemeester:
LGTM 🐰

duglin:
Anyone else finding this separate repo really annoying yet? :-)
Seeing this in isolation (and w/o the normal docker tests), and having to coordinate PRs across two repos, is **not** better than just making this a package under docker/docker.
See https://github.com/docker/docker/pull/19746 for another example.

duglin:
LGTM

but please do not merge until docker/docker#17329 is ready to be merged so we know for sure we want it and everything works ok.

dnephin:
Why would the api need to know about this flag? Can't this quiet be implemented entirely on the client side (by not printing to stdout)? Seems unnecessary to make it part of the interface.

I also just realized that our "api definition" seems to be completely missing query params? If that's correct, that's really unfortunate. A large part of the API is not defined by types at all.

> Anyone else finding this separate repo really annoying

maybe, but I think this shows that it's valuable for catching potential issues like above.

duglin:
REST API users should be able to turn this off too.

duglin:
not sure which issue you're referring to but if its about missing docs then I don't see why a separate repo is needed to expose that.  Everything in this repo could be under a pkg in docker/docker and any specialized testing we have here could become unit tests over there.

Given go's ability to import certain dirs from other repos I see no benefit but lots of pain with this approach.

dnephin:
> REST API users should be able to turn this off too.

Isn't "off" just "don't read bytes from the body of the response" ? Why would you need a flag for that?

> not sure which issue you're referring to 

The issue I raised about the new api param being unnecessary.

> I don't see why a separate repo is needed to expose that.

Definitely not necessary, I just think it makes it easier to review changes to the API. By making it a separate repo the API is a first-class concern.  Mixing in API changes with code changes can sometimes hide problems with the API (or at least make them harder to spot).

duglin:
On Jan 28, 2016, at 12:40 PM, Daniel Nephin <notifications@github.com> wrote:
> 
> REST API users should be able to turn this off too.
> 
> Isn't "off" just "don't read bytes from the body of the response" ? Why would you need a flag for that?
> 
a client may not want the noise at all on their networks.

> not sure which issue you're referring to
> 
> The issue I raised about the new api param being unnecessary.
> 
> I don't see why a separate repo is needed to expose that.
> 
> Definitely not necessary, I just think it makes it easier to review changes to the API. By making it a separate repo the API is a first-class concern. Mixing in API changes with code changes can sometimes hide problems with the API (or at least make them harder to spot).
> 
As of now one, for the most part, one can’t happen w/o the other - meaning we can’t fully test API changes w/o the code behind it to see how it will really look-n-feel.  It might be nice to claim that the API is independent of the engine but the reality is that, as of now, its not and except for trivial bug fixes any change in this repo will most likely requite a corresponding change in the engine code. And that PR coordination is annoying, time consuming and a time waste.

duglin:
no real opinion on the flag name - we appear to do both (use short/letters and use full words) at times.

calavera:
This needs a rebase.

coolljt0725:
rebased

calavera:
LGTM

PR #71, opened 2016-02-03
Add experimental interface for Checkpoint/Restore.

boucher:
The interface is a prerequisite for C/R support in the docker engine.

Signed-off-by: Ross Boucher <rboucher@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,9 @@ deps:
 test:
 	go test -tags=test ./...
 
+test_experimental:
+	go test -tags="test experimental" ./...
+
 validate:
 	go vet ./...
 	test -z "$(golint ./... | tee /dev/stderr)"
--- a/client/container_checkpoint.go
+++ b/client/container_checkpoint.go
@@ -1,0 +1,15 @@
+// +build experimental
+
+package client
+
+import (
+	"github.com/docker/engine-api/types"
+)
+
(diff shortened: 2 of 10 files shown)


boucher:
The tests pass, but I'm fairly sure they're only running the non experimental build. Not sure if you want to add explicit tests/build support in here for experimental mode, or just let those tests be handled at the Docker level.

vdemeester:
@boucher I think we might want to add tests for experimental mode at some point, but I don't think this is the scope of this PR :stuck_out_tongue:

boucher:
Great :)

calavera:
it would be good if you could add a new Make target to run the tests with the experimental flag on in this PR, we can run it in travis with the normal suite. At least, we'd be certain that this compiles :)

boucher:
@calavera pardon my Go newness. Is something like this in the Makefile what I'm looking for?

    test_experimental:
        go test -tags="test experimental" ./...

estesp:
This looks good based on the fact it looks like a clean migration of the experimental API side of C/R to `engine-api`.  If we're fine with it, it would be great to get this in so we can re-vendor and get the rest of the C/R capability into `docker/docker`.

boucher:
@crosbymichael I'm closing this PR as well, based on our discussion.

crosbymichael:
@boucher ok, thanks!

PR #76, opened 2016-02-04
Fix implicit conversion of transport sender in canceler_14.

calavera:
Fix docker build with gccgo:

```
00:22:12 # github.com/docker/engine-api/client/transport/cancellable
00:22:12 vendor/src/github.com/docker/engine-api/client/transport/cancellable/cancellable.go:40:21: error: argument 1 has incompatible type (need explicit conversion; missing method 'Scheme')
00:22:12   cancel := canceler(client, req)
```

### DIFF
--- a/client/transport/cancellable/canceler_go14.go
+++ b/client/transport/cancellable/canceler_go14.go
@@ -16,7 +16,7 @@ type requestCanceler interface {
 	CancelRequest(*http.Request)
 }
 
-func canceler(client transport.Client, req *http.Request) func() {
+func canceler(client transport.Sender, req *http.Request) func() {
 	rc, ok := client.(requestCanceler)
 	if !ok {
 		return func() {}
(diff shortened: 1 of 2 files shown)


vdemeester:
@calavera wondering something (but not really in the scope of the PR). Isn't the build flag of `canceler_go14.go` wrong ? It is `// +build !go1.5`, so when building with go1.6 it's gonna be taking into account to right ? If true, should be `// +build go1.4`.

Otherwise LGTM 🐲

runcom:
> It is // +build !go1.5, so when building with go1.6 it's gonna be taking into account to right ? If true, should be // +build go1.4.

agree LGTM btw

vdemeester:
> @calavera wondering something (but not really in the scope of the PR). Isn't the build flag of `canceler_go14.go` wrong ? It is `// +build !go1.5`, so when building with go1.6 it's gonna be taking into account to right ? If true, should be `// +build go1.4`.

Nevermind, my assumption was wrong : "go1.5", from Go version 1.5 onward — https://golang.org/pkg/go/build/.

runcom:
:smile: :smile:

Issue #77, opened 2016-02-04
Swagger specs


ahmetalpbalkan:
It came up a lot in the Docker Remote API rework discussions: https://github.com/docker/docker/issues/15188 https://github.com/docker/docker/issues/5893 I also think swagger is a great way to declare things.

If there's a way maybe we can declare types/* in swagger specs and generate source files using swagger code generators, that would be very preferable as other language SDKs can take advantage of the swagger specs as well.

Just leaving this here.

dnephin:
Also: https://github.com/docker-php/docker-php/blob/1.21/docker-swagger.json

I agree it would be good to either generate the go types from the spec, or generate the spec from  the go code. Both can work, and we seem to have both sources already.

I think engine-api is missing a big part of the interface specification (the path, query params, headers, and response codes).  Using swagger would force us to include all of those (which I think is good).

MHBauer:
Good idea to get it standardized format.

Does swagger have something to handle the stream hijacking that is done for attach, etc?

ahmetalpbalkan:
Well, swagger supports plugins/extensions. If there is no way of specifying hijacking, docker can specify hijack mode and code generators would implement it generically. 

More generally, people don't use hijack/attach very often in the client libraries. I tend to believe it's not much of a common task. They often do container/image managament as expected.

dnephin:
> Does swagger have something to handle the stream hijacking that is done for attach, etc?

I've been thinking about this. I think the way to support it would be to use `produces: ['application/x-json-stream']` and the "response type" would be the format of each item in the stream. Clients would need to either support that mime-type or allow for plugins to support it.

calavera:
> I agree it would be good to either generate the go types from the spec, or generate the spec from the go code. Both can work, and we seem to have both sources already.

we already have the types, let's generate the spec from them. That's what kubernetes does, bth. We might be able to reuse their scripts.

casualjim:
fyi: http://goswagger.io/generate/spec/

joelwurtz:
Could also use this spec for auto testing, where the example part of the path can be the acceptance test.

sflxn:
[...]

PR #78, opened 2016-02-05
Add JSON schema for API types that return JSON.

calavera:
We could validate API responses if we had these schemas, but this is just a PoC.

- Add script to generate JSON schema.

### DIFF
--- a/schema/generator/schema-generator.go
+++ b/schema/generator/schema-generator.go
@@ -1,0 +1,99 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/calavera/go-jsonschema-generator"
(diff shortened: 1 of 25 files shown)


dnephin:
Why jsonschema instead of a more complete solution like swagger (which is a superset of jsonschema) (#77)?

The problem with jsonschema is that it ignores a significant part of the API. Only request/response body are covered right? It's missing path, query params, headers, response headers and response codes, all important parts of the API definition.

Instead of checking it in, wouldn't it make sense to have these generated as part of the build? That way everyone who makes a change isn't forced to do the extra step of generating them and adding the changes.  Was it checked in just to show off what they look like?

LK4D4:
I agree that swagger is better and also it's supported by Linux Foundation now.

calavera:
Agreed, thanks for checking!

PR #83, opened 2016-02-09
Omit empty auth value serializing configuration.

calavera:
Do not serialize auth credentials when the auth field is empty. This works with https://github.com/docker/docker/pull/20107 where we're making this field optional since credentials can be stored in an external store, like the OS X keychain.


vdemeester:
LGTM 🐸

jfrazelle:
LGTM

PR #84, opened 2016-02-10
Add userns to run flags

liron-l:
Following the discussions in #20111, #19995 and #17409 we need a
mechinshem to enable skipping userns flag when userns re-mapping was
specified in the daemon level.
As suggest by #mrunalp, a coherent and consistent options is to add the
userns flag to the engine-api.

I've also moved the UT found at:
/docker/docker/runconfig/hostconfig_test.go next to hostconfig.go file
(since they test the specified functionality).

### DIFF
--- a/types/container/host_config.go
+++ b/types/container/host_config.go
@@ -65,6 +65,30 @@ func (n IpcMode) Container() string {
 	return ""
 }
 
+// UsernsMode represents userns mode in the container.
+type UsernsMode string
+
+// IsHost indicates whether the container uses the host's userns.
+func (n UsernsMode) IsHost() bool {
+	return n == "host"
+}
+
+// IsPrivate indicates whether the container uses the a private userns.
+func (n UsernsMode) IsPrivate() bool {
+	return !(n.IsHost())
+}
+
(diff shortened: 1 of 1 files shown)


liron-l:
@calavera can you please direct me to the appropriate reviewer for this change?

calavera:
TBH, I'd rather not bring these types here if they're not used in both sides, client and server. This is very similar to https://github.com/docker/engine-api/pull/56.

liron-l:
@calavera your comments is about ```DefaultDaemonNetworkMode``` not about the ```UsernsMode``` right? 
Also, what about the unit test refactor, are you ok with it?

liron-l:
@calavera I've cleanup this change and removed all unrelated code.
The new ```UsernsMode``` should be used both in client and server.

liron-l:
@calavera @estesp can you please review this change?

estesp:
LGTM - this is in support of the new flag to allow a privileged mode run when user namespaces are enabled. I believe it is used on both client and server as the flag parsing requires these functions.

calavera:
LGTM

PR #88, opened 2016-02-13
Updating NetworkName to return proper name for user defined networks 

msabansal:
The network api doesn't return the proper network name for user defined networks and it causes issues with windows libnetwork integration. This fixes that and brings parity with Linux implementation.

### DIFF
--- a/types/container/hostconfig_windows.go
+++ b/types/container/hostconfig_windows.go
@@ -15,9 +15,38 @@ func (n NetworkMode) IsNone() bool {
 	return n == "none"
 }
 
+// IsContainer indicates whether container uses a container network stack.
+// Returns false as windows doesn't support this mode
+func (n NetworkMode) IsContainer() bool {
+	return false
+}
+
+// IsBridge indicates whether container uses the bridge network stack
+// in windows it is given the name NAT
+func (n NetworkMode) IsBridge() bool {
+	return n == "nat"
+}
+
+// IsHost indicates whether container uses the host network stack.
+// returns false as this is not supported by windows
+func (n NetworkMode) IsHost() bool {
+	return false
+}
+
+// IsPrivate indicates whether container uses it's private network stack.
+func (n NetworkMode) IsPrivate() bool {
+	return !(n.IsHost() || n.IsContainer())
+}
+
+// ConnectedContainer is the id of the container which network this container is connected to.
+// Returns blank string on windows
+func (n NetworkMode) ConnectedContainer() string {
+	return ""
+}
+
 // IsUserDefined indicates user-created network
 func (n NetworkMode) IsUserDefined() bool {
-	return !n.IsDefault() && !n.IsNone()
+	return !n.IsDefault() && !n.IsNone() && !n.IsBridge()
 }
 
 // IsHyperV indicates the use of a Hyper-V partition for isolation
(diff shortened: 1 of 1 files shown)


calavera:
LGTM

msabansal:
Can we please get this merged? I need to start my review for docker daemon changes after this.

thaJeztah:
@msabansal to speed up the process; the `vendor` CI check is a separate check in the https://github.com/docker/docker repository; you can include these changes as part of your PR in docker/docker (which will make the `vendor` CI go red, but all other tests still run) but the PR can still be reviewed in parallel. Once the changes here are merged, you can update the vendor in the docker/docker PR to make the `vendor` CI go green

msabansal:
@thaJeztah Thanks. Done that.

thaJeztah:
Great, I saw you opened the PR, thanks!

jhowardmsft:
ping @calavera @thaJeztah David/Sebastiaan - any chance we can get this merged so that we can proceed with the revendoring into docker/docker and the integration of HNS on Windows? Thanks :dog2:

types/container/hostconfig_windows.go
```diff
@@ -15,9 +15,38 @@ func (n NetworkMode) IsNone() bool {
...
+// IsContainer indicates whether container uses a container network stack.
+// Returns false as windows doesn't support this mode
+func (n NetworkMode) IsContainer() bool {
+	return false
+}
+
+// IsBridge indicates whether container uses the bridge network stack
+// in windows it is given the name NAT
+func (n NetworkMode) IsBridge() bool {
+	return n == "NAT"
```
jhowardmsft: Oooh - msabansal, should this be lowercase too?

thaJeztah: :+1: yes, I think that should be changed

thaJeztah:
LGTM once that string is changed

types/container/hostconfig_windows.go
msabansal: Doing this. This will however require a minor libnetwork change aswell. Making that aswell.

calavera:
merged! :tada:

calavera:
There have been some changes in engine-api that might make your life slightly more complicated than expected. We might want to wait until https://github.com/docker/docker/pull/20565 is merged.

jhowardmsft:
Thanks @calavera. @msabansal Over to you for the next one :)

types/container/hostconfig_windows.go
```diff
@@ -15,9 +15,38 @@ func (n NetworkMode) IsNone() bool {
 	return n == "none"
 }
 
+// IsContainer indicates whether container uses a container network stack.
+// Returns false as windows doesn't support this mode
+func (n NetworkMode) IsContainer() bool {
+	return false
+}
+
+// IsBridge indicates whether container uses the bridge network stack
```
stevvooe: From a networking perspective, this is one of the most confusing statements I have ever read. Why would `bridge` and `nat` be the same thing?

msabansal: How so? The windows equivalent of the bridge mode is nat and is enabled through the WinNAT service.

stevvooe: A bridge is a very specific network device that "bridges" layer two packets over an interface, creating a single, layer 2 broadcast domain. A `nat` actually translates IP addresses between two separate networks, operating mostly at layer 3. There are protocols that work with one and not the other. For example, standard DHCP might run fine over a bridge but would require a relay or helper to work across a NAT boundary.

It sounds like windows doesn't really have the concept of a `bridge`. It would be better to leave out the concept of bridge than to have something claim it is a bridge but isn't.

I'm not privy to the implications of this behavior here, but it seems unwise to blur these concepts, as they are _not_ at all the same. Even if they are two approaches to solve the same problem in docker, they have different trade offs and behavior.

PR #92, opened 2016-02-16
Add unit tests for ImageHistory

vdemeester:
It's time for `ImageHistory` tests 🐮.

🐸

### DIFF
--- a/client/image_history_test.go
+++ b/client/image_history_test.go
@@ -1,0 +1,60 @@
+package client
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/docker/engine-api/client/transport"
+	"github.com/docker/engine-api/types"
+)
+
+func TestImageHistoryError(t *testing.T) {
+	client := &Client{
+		transport: transport.NewMockClient(nil, transport.ErrorMock(http.StatusInternalServerError, "Server error")),
+	}
+	_, err := client.ImageHistory("nothing")
+	if err == nil || err.Error() != "Error response from daemon: Server error" {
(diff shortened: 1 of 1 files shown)


duglin:
Testing just an API definition in isolation is hard and I know this is just following the pattern of other tests, but I'm not sure I see the value of tests that are really just testing that can serialize things properly.  If we assume that the core transport-level infrastructure is correct, then what specifically about the ImageHistory code could fail in these tests?  It seems like the only thing that might be an issue is if our mock data was incorrect - which means we're testing our testing, not the real code.

Let's not add tests that give a false sense of security. Let's focus on testing code that has "logic" in it - like the util funcs in the code, or (more importantly) that the API definition doesn't break the main implementation that matters.

duglin:
Thinking about this more.... I think it might be better if instead of testing the serialization code, or transport layer code, which we should test but not under the scope of a specific API, we should instead test that our client API code itself works.  Meaning, each API take a set of params (for example, ImageHistory takes just the ID/name of the container), let's verify that our code (as simple as it is) works correctly.  This means our test should verify that the URL produced matches the pattern defined in the API definition - ie. /images/ID/history.  Likewise, for other commands we could verify that each variable appears in the URL in the right spot - as part of the path vs as a query param (and with the right query key). 

I think this might have a better chance of catching a bug in the ImageHistory code - e.g. if someone changed a query param key by mistake.

duglin:
To be clear, I think keeping the existing tests in this PR is fine, but let's also add some for the in-bound side.

vdemeester:
> This means our test should verify that the URL produced matches the pattern defined in the API definition - ie. /images/ID/history. 

@duglin true :+1: I'll update these (and the other tests) to add the check on the `URL`.

calavera:
LGTM

LK4D4:
LGTM

PR #97, opened 2016-02-18
strslice: refactor StrSlice to avoid unnecessary pointer traversal

stevvooe:
By removing the struct wrapper around StrSlice, we maintain the slice
properties of the type (len, cap, make, etc.) while overriding the
serialization. We get a simpler type with less pointer traversal.

### DIFF
--- a/types/container/config.go
+++ b/types/container/config.go
@@ -24,12 +24,12 @@ type Config struct {
 	OpenStdin       bool                  // Open stdin
 	StdinOnce       bool                  // If true, close stdin after the 1 attached client disconnects.
 	Env             []string              // List of environment variable to set in the container
-	Cmd             *strslice.StrSlice    // Command to run when starting the container
+	Cmd             strslice.StrSlice     // Command to run when starting the container
 	ArgsEscaped     bool                  `json:",omitempty"` // True if command is already escaped (Windows specific)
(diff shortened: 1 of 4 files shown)


calavera:
LGTM

LK4D4:
LGTM

runcom:
This has broken `go get` for ppl using docker :3

stevvooe:
@runcom How?

runcom:
@stevvooe go get a project which uses docker image/ library for instance is broken because Cmd has no Len() method anymore here in engine-api while docker is still using it

stevvooe:
@runcom Unfortunately, we had to break interface compatibility. This was technical debt that had to be addressed. To avoid problems with `go get`, we typically suggest that projects should use vendoring.

In the future, we can avoid this by keeping the standards of review higher.

runcom:
Again, this has nothing to do with technical debt, I'm not sure why we're still saying this. This was about a strictly backward compatible change that had to be done. Now what's happening is that this going to break that compatibility, again.

runcom:
Integration and unit test against docker/docker do not lie. That's the format expected even if it wasn't the *good* one,it was just the right one at the time and with this, tests fail. Full stop. Nothing to do with review processes.

PR #98, opened 2016-02-19
Correctly set content type for image load

estesp:
The default content type header is `text/plain` if none is specified, which, if authz plugins are in
use will cause the [entire body to be read](https://github.com/docker/docker/blob/master/pkg/authorization/authz.go#L150-L152) (and fail because of another
bug that will be fixed in **docker/docker** separately). Given the content
type should really be binary format anyway, this change should be useful for
any other code in the future that might determine actions based on POST
`Content-Type` headers.

Docker-DCO-1.1-Signed-off-by: Phil Estes <estesp@linux.vnet.ibm.com> (github: estesp)

### DIFF
--- a/client/image_load.go
+++ b/client/image_load.go
@@ -18,7 +18,8 @@ func (cli *Client) ImageLoad(ctx context.Context, input io.Reader, quiet bool) (
 	if quiet {
 		v.Set("quiet", "1")
 	}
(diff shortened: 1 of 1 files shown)


estesp:
See docker/docker#20508 for more information on the additional issues with POST body pass-through in the authz plugin framework

LK4D4:
LGTM

vdemeester:
LGTM 🐰

PR #104, opened 2016-02-21
Split volume.go into multiple files

vdemeester:
To follow what we do with containers and images 🐯.

🐸


calavera:
LGTM

PR #107, opened 2016-02-23
Change Networks to NetworksStats for container stats

wenchma:
Related to docker/docker#20596

Signed-off-by: wenchma <wenchma@cn.ibm.com>

### DIFF
--- a/types/stats.go
+++ b/types/stats.go
@@ -108,5 +108,5 @@ type StatsJSON struct {
 	Stats
 
 	// Networks request version >=1.21
-	Networks map[string]NetworkStats `json:"networks,omitempty"`
+	Networks map[string]NetworkStats `json:"networks_stats,omitempty"`
 }


calavera:
I rather not do this. It's a breaking change in the api that we'll need to support as backwards compatible.

thaJeztah:
Not convinced either at this point; sure it's slightly nicer, but not sure it's worth the hassle

wenchma:
@calavera @thaJeztah  understanding,  if it's like that :

    Networks map[string]NetworkStats `json:"networks_stats,omitempty"`

just update the json output, is it accepted ?

calavera:
No. Calls to old API versions still need to keep the old field name, to be backwards compatible. That means that you have to fork the stats field by API version only to call the same field `networks_stats` for newer versions and `networks` for older versions. I think adding this extra logic is not worth it.

calavera:
I see the issue is closed already. I'm closing this PR too. Thanks a lot for your clear description in the issue.

PR #108, opened 2016-02-23
Add support for getting token from login

dmcgowan:
The auth response will return an identity token from a login when long lived token authentication is supported. The identity token can be used in the auth config in place of basic authentication to get registry access tokens.

See https://github.com/docker/distribution/pull/1418 for further context on supporting long lived oauth tokens.

### DIFF
--- a/types/auth.go
+++ b/types/auth.go
@@ -7,5 +7,11 @@ type AuthConfig struct {
 	Auth          string `json:"auth,omitempty"`
 	Email         string `json:"email"`
 	ServerAddress string `json:"serveraddress,omitempty"`
+
+	// IdentityToken is used to authenticate the user and get
+	// an access token for the registry.
+	IdentityToken string `json:"identitytoken,omitempty"`
+
+	// RegistryToken is a bearer token to be sent to a registry
 	RegistryToken string `json:"registrytoken,omitempty"`
 }
--- a/types/types.go
+++ b/types/types.go
@@ -39,6 +39,10 @@ type ContainerUpdateResponse struct {
 type AuthResponse struct {
 	// Status is the authentication status
 	Status string `json:"Status"`
+
+	// IdentityToken is an opaque token used for authenticating
+	// a user after a successful login.
+	IdentityToken string `json:"IdentityToken,omitempty"`
 }
 
 // ContainerWaitResponse contains response of Remote API:


MHBauer:
Why `identitytoken` on one and `IdentityToken` on the other? (Capitalization differences)

dmcgowan:
@MHBauer trying to be consistently inconsistent. Not sure why the responses use CamelCase but the auth type doesn't and I would rather match the case of the field structs.

calavera:
LGTM although I'd rather wait until we have a PR in docker/docker to handle this before merging it.

dmcgowan:
@calavera I have no problem waiting to merge

dmcgowan:
We are ready to open up a PR on docker/docker now that https://github.com/docker/distribution/pull/1475 has been merged.

aaronlehmann:
LGTM

@calavera: Are you okay with merging this now, with the Docker Engine PR coming later today? Then the engine PR will be able to vendor a real `engine-api` update instead of a fake one.

calavera:
Merged! :tada:

aaronlehmann:
@calavera: Thanks!

types/auth.go
```diff
@@ -7,5 +7,11 @@ type AuthConfig struct {
 	Auth          string `json:"auth,omitempty"`
 	Email         string `json:"email"`
 	ServerAddress string `json:"serveraddress,omitempty"`
+
+	// IdentityToken is used to authenticate the user and get
+	// an access token for the registry.
+	IdentityToken string `json:"identitytoken,omitempty"`
```
icecrime: Sorry I'm late to this party, but I do feel that @calavera comment in https://github.com/docker/docker/pull/20970#issuecomment-192535355 does seem to make sense. Do we need a new field if `IdentityToken` and `Password` are mutually exclusive? I wouldn't fight for it much if there's a nice benefit to do it this way. Thanks!

dmcgowan: They are mutually exclusive but either way a field would be required in order to differentiate the secret type. For storage of the auth config I think it is fine to combine the "secret" information into a single field which could be the focus of protection, but at this point in the code having a fully enumerated structure is in my opinion the better design.

icecrime: Thanks @aaronlehmann for clarifying for me: `Password` is sent with basic auth, but `IdentityToken` goes over `Authorization: Bearer ...`, so the client needs to know which one it is dealing with.

PR #109, opened 2016-02-24
Add IO resource control settings for Windows

darrenstahlmsft:

### DIFF
--- a/types/container/host_config.go
+++ b/types/container/host_config.go
@@ -166,6 +166,7 @@ type LogConfig struct {
 type Resources struct {
 	// Applicable to all platforms
 	CPUShares int64 `json:"CpuShares"` // CPU shares (relative weight vs. other containers)
+	Memory    int64 // Memory limit (in bytes)
 
 	// Applicable to UNIX platforms
 	CgroupParent         string // Parent cgroup.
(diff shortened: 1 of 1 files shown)


types/container/host_config.go
```diff
@@ -181,13 +182,17 @@ type Resources struct {
...
 	KernelMemory         int64           // Kernel memory limit (in bytes)
-	Memory               int64           // Memory limit (in bytes)
 	MemoryReservation    int64           // Memory soft limit (in bytes)
 	MemorySwap           int64           // Total memory usage (memory + swap); set `-1` to enable unlimited swap
 	MemorySwappiness     *int64          // Tuning container memory swappiness behaviour
 	OomKillDisable       *bool           // Whether to disable OOM Killer or not
 	PidsLimit            int64           // Setting pids limit for a container
 	Ulimits              []*units.Ulimit // List of ulimits to be set in the container
+
+	// Applicable to Windows platforms
```
jhowardmsft: Just Windows :smile:

jhowardmsft:
Minor nit, otherwise LGTM

types/container/host_config.go
darrenstahlmsft: Fair

calavera:
@jhowardmsft I wonder if we should have two different structures by platform. We've talked about this in the past and it might be a good moment to do this, since most of the fields in the `Resources` struct don't apply to both. Is it a good idea to have a `CommonResources` with shared fields and a `Resources` struct by platform?

jhowardmsft:
@calavera Unfortunately we can't. I had a PR that did this a long time back, but realised it wasn't workable. The reason is that this is a common structure passed from client to daemon. The client has no idea whether the daemon is Windows or Linux, so has to pass all fields. Otherwise a Windows client wouldn't be able to correctly communicate with a Linux daemon, or visa-versa.

calavera:
> The client has no idea whether the daemon is Windows or Linux, so has to pass all fields

:pensive: :penguin: 

LGTM

PR #113, opened 2016-02-25
Removing dead code duplicated in docker/docker

darrenstahlmsft:
As far as I can tell, these functions are duplicated in docker/docker and the docker/docker version is used in all cases. I can submit a parallel PR to docker/docker to confirm that CI passes if needed, but it builds and runs fine locally with these changes.

### DIFF
--- a/types/container/hostconfig_windows.go
+++ b/types/container/hostconfig_windows.go
@@ -1,7 +1,6 @@
 package container
 
 import (
-	"fmt"
 	"strings"
 )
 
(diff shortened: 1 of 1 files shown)


MHBauer:
I concur it is not used. I wonder if the usage in `runconfig` will be moving here with the server side or not. It is duplicate code, and if it comes back so be it.

msabansal:
There is another PR pending https://github.com/docker/engine-api/pull/88/ which is required for windows libnetwork integration. NetworkName will be required once Windows libnetwork integration happens(refer https://github.com/docker/docker/pull/20478).

darrenstahlmsft:
Updated to not remove NetworkName

calavera:
LGTM but it needs a rebase.

calavera:
:fire: dead code :fire: :exclamation:

PR #133, opened 2016-03-06
Add `AutoRemove` into HostConfig

WeiZhang555:
See https://github.com/docker/docker/pull/20848#issuecomment-191882980

Docker will move `--rm` flag from client side to daemon side, this
demands HostConfig should have a flag for `--rm`.

### DIFF
--- a/types/container/host_config.go
+++ b/types/container/host_config.go
@@ -239,6 +239,7 @@ type HostConfig struct {
 	NetworkMode     NetworkMode   // Network mode to use for the container
 	PortBindings    nat.PortMap   // Port mapping between the exposed port (container) and the host
 	RestartPolicy   RestartPolicy // Restart policy to be used for the container
+	AutoRemove      bool          // Automatically remove container when it exits
 	VolumeDriver    string        // Name of the volume driver used to mount volumes
 	VolumesFrom     []string      // List of volumes to take from other container
 


cpuguy83:
Only concern here is having a bool vs a string that can potentially have other values rather than a strict yes/no.
For instance, `--rm on-success`

thaJeztah:
@cpuguy83 yes, I raised that concern on the PR https://github.com/docker/docker/pull/20848#issuecomment-191847173, current idea is to add an extra `AutoRemoveOn` (e.g.) property if we want to add more options (so `AutoRemove=true`, `AutoRemoveOn=success`). But I'm still open to the alternative if more people think that's the better option.

WeiZhang555:
@cpuguy83 Yes, I'm just about to comment, but @thaJeztah said it all :smile:

calavera:
LGTM

amatsus:
We run a container with --rm option:

```
$ docker run --rm --name nginx-test nginx
```
and in another terminal, run:

```
$ docker inspect nginx-test | grep AutoRemove
            "AutoRemove": false,
```
Is this value correct?

WeiZhang555:
@amatsus It's not.

How can this happen? I can't reproduce following your steps.

amatsus:
@WeiZhang555 You see the same result in the following comments:
https://github.com/docker/docker/issues/22938#issue-156474634
https://github.com/docker/docker/issues/22684#issue-154411517

and I get this with the following docker version:

```
$ echo `docker --version` on `grep -Po '(?<=PRETTY_NAME=")(.*)(?=")' /etc/os-release`
Docker version 1.12.1, build f1040da/1.12.1 on CentOS Linux 7 (Core)

$ echo `docker --version` on `grep -Po '(?<=PRETTY_NAME=")(.*)(?=")' /etc/os-release`
Docker version 1.11.2, build 3405dd7/1.11.2 on CentOS Linux 7 (Core)

$ echo `docker --version` on `grep -Po '(?<=PRETTY_NAME=")(.*)(?=")' /etc/os-release`
Docker version 1.11.0, build 4dc5990 on Ubuntu 14.04.4 LTS

$ /Applications/Docker.app/Contents/Resources/bin/docker-diagnose| head -2
OS X: version 10.11.6 (build: 15G31)
Docker.app: version: 1.12.1-beta24 (0afa5e5)
```
I hope this helps.

cpuguy83:
@amatsus It's because this was merged way before the feature got into docker.

amatsus:
@cpuguy83 Thank you for your information. I get value of AutoRemove with docker 1.13.0-dev. I look forward to this feature release.

PR #136, opened 2016-03-07
Implement Jenkins jobs using DSL

mikedougherty:
This creates a set of jobs to test this repo in various ways: vet, lint,
fmt, cyclo, unittests (using gocov test) on linux and windows. It also
includes jobs to trigger certain git refs: tags, branches, PRs, and the
default branch (master).

Cyclomatic complexity max is set to 21 and should be reduced over time as
functions are refactored.

vet, lint, fmt, and cyclo are all recorded using the Warnings plugin and
are currently set to fail the job if *any* issues are detected. Another possible
way to do this is set the maximum number of problems to 2 and unset the
cyclomatic complexity max. This way the 2 cyclo issues won't be hidden
away (and adding another 19 complexity function would fail the job).

Test and coverage results are recorded in a jenkins-friendly fashion
using JUnit and Cobertura plugins respectively.

----
All of the jobs should go "yellow" when warnings or test failures happen, which this should make it more obvious when there are code issues vs problems evaluating the code.

You can see an example of the results recorded in jenkins [here](https://jenkins.dockerproject.org/job/mikedougherty-engine-api/job/task-test-windows/), or a screenshot [here](https://cloud.githubusercontent.com/assets/1046019/13582746/e88b7394-e462-11e5-9936-5294a1ac831e.png). There are of course plenty of clickable things and ways to drill down into failures.

[This PR](https://github.com/mikedougherty/engine-api/pull/1) also shows the status labels that will be used, which can be amended if desired. I believe they don't show on this PR due to fork/permissions complications.

I'd like to see this tool used in more projects as you can see it enables us to source control the CI configuration as well as simplifying changes across many jobs at once. Over time I'll be working on a library which will make some of the more common configurations easier or enabled by default. Currently some of this will need to be duplicated/altered into other projects (undesirable in the long run).

Another piece of the puzzle not yet implemented is generating a job which will watch the repository for changes to the DSL script itself and update the jobs accordingly. Currently I've created this manually and you can see it [here](https://jenkins.dockerproject.org/job/dsl/job/mikedougherty-engine-api/).

cc: @jhowardmsft @icecrime @SvenDowideit @calavera

### DIFF
--- a/hack/test-linux.sh
+++ b/hack/test-linux.sh
@@ -1,0 +1,11 @@
+#!/bin/bash -x
+docker pull dockerautomation/golang-tester:gimme
+docker run \
+--rm \
+-v "$(pwd):/go/src/$GOPACKAGE" \
+-v "$(pwd)/results:/output" \
+-e "GOVERSION=$GOVERSION" \
+-e "GOCOV_ARGS=-race -tags=test" \
+-e "GOCYCLO_MAX" \
+-e "GOPACKAGE" \
+dockerautomation/golang-tester:gimme
--- a/hack/test-windows.sh
+++ b/hack/test-windows.sh
@@ -1,0 +1,35 @@
+#!bash -x
+mkdir -p "/tmp/$BUILD_TAG"
+trap "rm -rf /tmp/$BUILD_TAG" EXIT
+
+export GOVERSION=1.5.3
+export GOPATH="/tmp/$BUILD_TAG"
+mkdir -p "$GOPATH/src/$GOPACKAGE"
+rm -rf "$GOPATH/src/$GOPACKAGE"
+cp -R "$WORKSPACE" "$GOPATH/src/$GOPACKAGE"
+
+export GOBIN="$GOPATH/bin"
+export PATH="$PATH:$GOBIN"
+export GOCOV_ARGS="-race -tags=test"
+
+mkdir -p "$GOBIN"
+curl -sSLo "$GOBIN/golang_test.sh" "https://${GITHUB_TOKEN}@raw.githubusercontent.com/docker/tools-team/master/Dockerfiles/golang-tester/entrypoint.sh"
+chmod u+x "$GOBIN/golang_test.sh"
+
+## Not using gimme currently
+# gimme "$GOVERSION"
+# curl -sSLo "$GOBIN/gimme "https://${GITHUB_TOKEN}@raw.githubusercontent.com/travis-ci/gimme/master/gimme
+# chmod u+x "$GOBIN/gimme"
+cp "$(which true)" "$GOBIN/gimme"
+
+go version
+
+go get github.com/golang/lint/golint
+go get github.com/fzipp/gocyclo
+go get github.com/axw/gocov/gocov
+go get github.com/AlekSi/gocov-xml
+go get bitbucket.org/tebeka/go2xunit
+
+mkdir -p results
+export OUTPUT_DIR="results"
+exec bash -x golang_test.sh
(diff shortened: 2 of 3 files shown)


mikedougherty:
Heh. we can also add dco stuff to the build if you like :)

calavera:
what's the benefit of putting it in our infra when public CI systems work fine for repos like this one? I personally don't care about cyclomatic complexity, tbh. I'd like to know if other people find it useful for any real work.

mikedougherty:
Mostly that we'd like to have a consistent way to do CI across projects, instead of scattering across various services we can consolidate and standardize expected tasks, results, views, access, and platforms.

This has few dependencies on other docker projects and we wanted to do testing on windows, so this was just a good target for a first attempt.

(/cc @dmp42)

---
re: cyclo, i included it since i saw it used on goreportcard.com. it's easy enough to include or remove, i opted to include it as an example. and who doesn't love more graphs? :)
personally i have found it helpful in previous projects as a signal to refactor a bit of code but it's by no means the only tool for that nor is it authoritative. it's just another actionable quantitative measure of a codebase.

MHBauer:
I have only seen cyclomatic complexity noted after the fact. It was mainly ignored, as it pointed out big complex methods that everyone knew about. I can see using it going forward, but only as a 'nice to have'. Some things are complex because they are complicated and some are complex because they are poorly written.

If this is the way all the projects are to be taken, I do not have any problem with it. It does become a single point of failure. I can't see anything directly in jenkins, but the screenshot looks okay. I would like if the results could be directly embedded here in github to avoid the extra click over to jenkins.

dmp42:
As a side note: is there anyway not to land a bunch of code in repos? (that will have to be duplicated and maintained).
Maybe we should sit together and discuss this more in-depth, as a common effort for all projects?

mikedougherty:
@dmp42 that's the goal; i would love to make something nice and neat and easy but everybody wants things yesterday so ugly things like 250+ lines of groovy happen.

---
re:cyclo it's really the least important part of this PR; hopefully it's obvious that it can be enabled/disabled/ignored as desired. i don't think it's worth spending time on. if it helps you can think of it as something equally unimportant like spellcheck :)

dmp42:
@mikedougherty I don't think there is any emergency to do things for yesterday :-)
Let's discuss this tomorrow.

jfrazelle:
ooo coool you are doing it \o/ its so prettty

On Mon, Mar 14, 2016 at 5:26 PM, Olivier Gambier <notifications@github.com>
wrote:

> @mikedougherty <https://github.com/mikedougherty> I don't think there is
> any emergency to do things for yesterday :-)
> Let's discuss this tomorrow.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/docker/engine-api/pull/136#issuecomment-196584942
>



-- 


Jessie Frazelle
4096R / D4C4 DD60 0D66 F65A 8EFC  511E 18F3 685C 0022 BFF3
pgp.mit.edu <http://pgp.mit.edu/pks/lookup?op=get&search=0x18F3685C0022BFF3>

friism:
@dmp42 to me, this work has some urgency because it unblocks other teams that want to start Windows CI testing in a good way

calavera:
> this work has some urgency because it unblocks other teams that want to start Windows CI testing in a good way

that's not relevant to this repo, we already have good enough windows testing here.

crosbymichael:
Have have moved this code back into the main docker repo so this will not be needed.  We are not deleting this but will not be making changes here.

PR #140, opened 2016-03-08
client: plumb net/context.Context through entire API

### LINKED ISSUE #139: Inconsistent use of context for APIClient methods
When using `APIClient.ContainerCreate`, I found that no context argument is present: https://github.com/docker/engine-api/blob/master/client/container_create.go#L21. This ends up using [`(*Client).post`](https://github.com/docker/engine-api/blob/master/client/request.go#L42), which uses the background context.

This basically makes the integration of context useless.

stevvooe:
In general, `net/context.Context` should be the first argument to
methods that may be on a request call chain. While some API methods had
context, many did not, making support inconsistent. Applications that
require full plumbing of context, such as those that rely on
cancellation for resource cleanup, would be incompatible with
inconsistent plumbing. The change takes a first pass at adding context
through the entire API.

Close #139

### DIFF
--- a/client/container_attach.go
+++ b/client/container_attach.go
@@ -4,13 +4,14 @@ import (
 	"net/url"
 
 	"github.com/docker/engine-api/types"
+	"golang.org/x/net/context"
 )
 
 // ContainerAttach attaches a connection to a container in the server.
 // It returns a types.HijackedConnection with the hijacked connection
 // and the a reader to get output. It's up to the called to close
 // the hijacked connection by calling types.HijackedResponse.Close.
-func (cli *Client) ContainerAttach(options types.ContainerAttachOptions) (types.HijackedResponse, error) {
+func (cli *Client) ContainerAttach(ctx context.Context, options types.ContainerAttachOptions) (types.HijackedResponse, error) {
 	query := url.Values{}
 	if options.Stream {
 		query.Set("stream", "1")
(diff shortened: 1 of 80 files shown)


vdemeester:
LGTM 🐮

thaJeztah:
ping @calavera :innocent:

stevvooe:
@thaJeztah @calavera @vdemeester This PR will have implications for the engine but the results will be fruitful. I hope that is okay.

calavera:
I didn't add that argument to all of them to isolate only the calls that are more susceptible of cancellation, but being consistent sounds great.

LGTM but you'll have to rebase it :stuck_out_tongue:

stevvooe:
@calavera Rebased!

aaronlehmann:
This needed to be done and I highly support the change.

...but I think invasive changes like this one should be accompanied by a PR to docker/docker. Otherwise it places a big burden on the next person who needs to vendor engine-api to pull in their own change. Their PR would become far larger and their actual change would get buried among widespread changes to get the engine compiling again.

stevvooe:
@aaronlehmann I mentioned that above. I'm not sure what the right approach here is.  If this is really a huge problem, we can automate the rewrite.

thaJeztah:
libcompose and Swarm are also switching to engine-api, so may need changes as well

PR #141, opened 2016-03-08
client: remove need for "test" tag to run tests

stevvooe:
Because client tests imported a test file from another package, `go
test` would not work without adding `-tags=test`. This is not the
correct invocation for regularly testing Go code and breaks simple runs
of `go test` working for the package. This moves the transport mock into
the package that actually uses it. Now, `go test` works properly in the
client package.


vdemeester:
@stevvooe @calavera I kinda like that (I'm currently hacking around in my editor to run tests) but I wonder if we could make this `mockClient` available from outside (for example to be used for unit tests in libraries that uses `engine-api`).

stevvooe:
@vdemeester The behavior in my editor was the forcing function of this PR. I think it could be made available but no reason to unless it's used, so I'd like to avoid it for this PR.

vdemeester:
@stevvooe seems good to me ;-)
LGTM 🐯
/cc @calavera

calavera:
LGTM

PR #152, opened 2016-03-16
ImageImport should use postRaw instead of post.

vdemeester:
Using post instead of postRaw makes client hang for ever. 

This is related to docker/docker#21219 🐳.

/cc @calavera @stevvooe @LK4D4 @cpuguy83 

🐸

### DIFF
--- a/client/image_import.go
+++ b/client/image_import.go
@@ -21,7 +21,7 @@ func (cli *Client) ImageImport(ctx context.Context, options types.ImageImportOpt
 		query.Add("changes", change)
 	}
 
-	resp, err := cli.post(ctx, "/images/create", query, options.Source, nil)
+	resp, err := cli.postRaw(ctx, "/images/create", query, options.Source, nil)
 	if err != nil {
 		return nil, err
 	}


vdemeester:
```
go vet ./...

client/hijack.go:138: assignment copies lock value to c: crypto/tls.Config contains sync.Once contains sync.Mutex
```

Hum something change on go devel :sweat:

calavera:
I opened a PR to fix it https://github.com/docker/engine-api/pull/154

calavera:
LGTM

MHBauer:
Do we understand why it hangs? My cursory read of the code shows the the headers are different, content type is text instead of json. Does that change the server response to streaming instead of one-shot or something?

Looks like it just un-does the previous change to the file.

vdemeester:
@MHBauer yeah : https://github.com/docker/engine-api/blob/075a1feac74ea2a2180a5b5b3ec747c2cdb56a82/client/request.go#L59

`post` assumes the body that is passed is not an `io.Reader` but an actual data sturct. So it tries to encode the thing and well.. simply put, with a `io.Reader` it hangs. The previous change was wrong on that particular file (but was part of more general changes).

stevvooe:
@vdemeester This one is my fault. It would be nice to get a little type-safety here, but there may be cases where we want the json serialization of a type that implements `Read`, albeit there may actually be no cases of this.

Typically, when building API clients, I use a pattern like this:

```
var (
    resp MyObject
    req PostObject 
    values url.Values
)

client.postObject(&resp, req, values)
```

It makes it explicit that we are sending and receiving objects and not readers.

The other option here might be to have an internal interface for types accepted by that method:

```
type apiObject {
    apiObject()
}

func (cli *Client) post(ctx context.Context, body apiObject,...) error
```

Then, anything being passed as a body must implement that method, ensuring we never pass anything that wasn't intended to go there. With a growing API, this might be a reasonable measure.

PR #162, opened 2016-03-20
Make required arguments required.

### LINKED ISSUE #137: Make required arguments required - optional arguments should be optional
It makes no sense to have the primary arguments for API operations inside of an options structure. It makes the API unwieldy and challenging to use. Let's take the example of [`Client.ImagePull`](https://godoc.org/github.com/docker/engine-api/client#Client.ImagePull):

```
func (cli *Client) ImagePull(ctx context.Context, options types.ImagePullOptions, privilegeFunc RequestPrivilegeFunc) (io.ReadCloser, error)
```

We then dive into [types.ImagePullOptions](https://godoc.org/github.com/docker/engine-api/types#ImagePullOptions). Ok, it takes an `ImageID`. Does that mean name? Who knows. It also requires the user to parse the tag portion out of the name.

Furthermore, the only real option, `privilegeFunc` is for some reason not an option.

The method signature should be this:

```
func (cli *Client) ImagePull(ctx context.Context, ref string, options types.ImagePullOptions) (io.ReadCloser, error)
```

Note a few things about this design. The `ref` can contain whatever is supported by the target engine. We don't require a user of this API to parse out a tag, which is non-trivial. Futhermore, one can simply pass an empty options structure and 95% of use cases would be covered.

With the options type, as follows:

```
type ImagePullOptions struct {
    RegistryAuth string // RegistryAuth is the base64 encoded credentials for the registry
    PrivilegesFunc RequestPrivilegesFunc
}
```

As it is, this is confusing and challenging to use correctly.

I'm not sure if I have time to do a PR for this, as it seems to be a systematic inconsistency.

vdemeester:
Change method signatures to make required arguments actually required (using a paramater), all accross the API 🐮.

*It's not complete (there is a few `FIXME` still), I'm opening the PR to gather opinion on this* ; I also think it should not be merged until docker `1.11` is released 🐵.

@stevvooe is this something that you had in mind ?

/cc @calavera @MHBauer @dongluochen 

This is related to and fixes #137.

🐸


types/client.go
```diff
@@ -184,40 +180,43 @@ type ImageLoadResponse struct {
 
 // ImagePullOptions holds information to pull images.
 type ImagePullOptions struct {
-	ImageID      string // ImageID is the name of the image to pull
-	Tag          string // Tag is the name of the tag to be pulled
-	RegistryAuth string // RegistryAuth is the base64 encoded credentials for the registry
+	Tag           string // Tag is the name of the tag to be pulled
```
stevvooe: Tag should be part of argument. For example, "redis:latest" should be part of the argument. For now, we have to parse this on the client side, but in the future, this format should be daemon specific.

client/interface.go
```diff
@@ -15,25 +15,25 @@ import (
 // APIClient is an interface that clients that talk with a docker server must implement.
 type APIClient interface {
 	ClientVersion() string
-	ContainerAttach(ctx context.Context, options types.ContainerAttachOptions) (types.HijackedResponse, error)
-	ContainerCommit(ctx context.Context, options types.ContainerCommitOptions) (types.ContainerCommitResponse, error)
+	ContainerAttach(ctx context.Context, containerID string, options types.ContainerAttachOptions) (types.HijackedResponse, error)
+	ContainerCommit(ctx context.Context, containerID string, options types.ContainerCommitOptions) (types.ContainerCommitResponse, error)
 	ContainerCreate(ctx context.Context, config *container.Config, hostConfig *container.HostConfig, networkingConfig *network.NetworkingConfig, containerName string) (types.ContainerCreateResponse, error)
-	ContainerDiff(ctx context.Context, ontainerID string) ([]types.ContainerChange, error)
+	ContainerDiff(ctx context.Context, containerID string) ([]types.ContainerChange, error)
```
stevvooe: This option might be okay as `container`, since this takes a name or id.

client/interface.go
```diff
@@ -46,22 +46,22 @@ type APIClient interface {
...
-	ImagePush(ctx context.Context, options types.ImagePushOptions, privilegeFunc RequestPrivilegeFunc) (io.ReadCloser, error)
-	ImageRemove(ctx context.Context, options types.ImageRemoveOptions) ([]types.ImageDelete, error)
-	ImageSearch(ctx context.Context, options types.ImageSearchOptions, privilegeFunc RequestPrivilegeFunc) ([]registry.SearchResult, error)
+	ImagePull(ctx context.Context, ref string, options types.ImagePullOptions) (io.ReadCloser, error)
+	ImagePush(ctx context.Context, ref string, options types.ImagePushOptions) (io.ReadCloser, error)
+	ImageRemove(ctx context.Context, imageID string, options types.ImageRemoveOptions) ([]types.ImageDelete, error)
+	ImageSearch(ctx context.Context, term string, options types.ImageSearchOptions) ([]registry.SearchResult, error)
 	ImageSave(ctx context.Context, imageIDs []string) (io.ReadCloser, error)
-	ImageTag(ctx context.Context, options types.ImageTagOptions) error
+	ImageTag(ctx context.Context, imageID, repository, tag string, options types.ImageTagOptions) error
```
stevvooe: I think this should be `imageID, ref string`, since this is how we talk about it in the cli. One tags with `redis:2`, not `redis --tag 2`. Otherwise, we push a lot of complex tag parsing down on users of the API.

stevvooe:
@vdemeester This is exactly what I was talking about. This is a fantastic start!

client/interface.go
MHBauer: I agree. It's unfortunate that the repository/tag combination has really been abused to be base/version, but I have not seen any other uses that are not of that form.

MHBauer:
I like it. Makes using the api more understandable at a glance.

client/interface.go
stevvooe: @MHBauer We should change this here. Just because the uses have poor practice doesn't mean we need to encourage. We should moving the parsing to the internals of this function.

vdemeester: Question is then : Should we use the `reference` package to parse `ref` without changing the call to the server API, or should we update the server API too (thus making some related changes on `docker/docker` and such) ?

stevvooe: This is indeed complicated:

1. The argument should be `ref` of type `string`. Let's not impose use of the reference package on the user.
2. Parse the tag from the reference using the `reference` package.
3. Maybe have a plan to change the server API in the future, but this is out of scope.

vdemeester: Hum that's mainly what I was thinking about.. Go a little bit *confused* with the `reference` package as there is one in `docker/distribution` and one in `docker/docker` :stuck_out_tongue_closed_eyes:.

Reading docker/docker#18731 and thus docker/distribution#1277, I guess I should use the `docker/distribution`, am I right ?

I think, I'll make an issue on *the server API change in the future*, when this one will be merged :wind_chime:.

stevvooe: @vdemeester Yes, you should use the `docker/distribution` package. There may be some default behavior you'll need to reproduce from `docker/docker/reference`, but it is mostly straightforward.

calavera:
go for it!

> I also think it should not be merged until docker 1.11 is released

agreed

client/image_reference.go
```diff
@@ -0,0 +1,27 @@
+package client
```
calavera: should this be in `types`?

Should we have a more general package under `types/reference`? We've talked about this several times, maybe now it's the right moment to do it

vdemeester: @calavera ah yes, good idea :wink:

vdemeester:
@stevvooe updated with `reference` *handling*. I'm not sure I got all right but :stuck_out_tongue_closed_eyes:

calavera:
LGTM

types/reference/image_reference.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"fmt"
+
+	distreference "github.com/docker/distribution/reference"
+)
+
+// parseReference parses the given references and return the repository and
+// tag (if present) from it. If there is an error during parsing, it will
+// return an error.
+func Parse(ref string) (string, string, error) {
```
vdemeester: @stevvooe yes. I think I'm gonna work on `docker/docker` with vendoring this branch to look at the result and see if I did miss something :angel:

types/reference/image_reference.go
```diff
@@ -0,0 +1,28 @@
...
+	if err != nil {
+		return "", "", err
+	}
+
+	var tag string
+	if tagged, isTagged := distributionRef.(distreference.NamedTagged); isTagged {
+		tag = tagged.Tag()
+	}
+	// This is in order to support a reference like :
+	// tag@sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
```
vdemeester: Ah I see what you mean :)

client/interface.go
```diff
@@ -15,54 +15,54 @@ import (
 // APIClient is an interface that clients that talk with a docker server must implement.
 type APIClient interface {
 	ClientVersion() string
-	ContainerAttach(ctx context.Context, options types.ContainerAttachOptions) (types.HijackedResponse, error)
-	ContainerCommit(ctx context.Context, options types.ContainerCommitOptions) (types.ContainerCommitResponse, error)
+	ContainerAttach(ctx context.Context, container string, options types.ContainerAttachOptions) (types.HijackedResponse, error)
+	ContainerCommit(ctx context.Context, container string, options types.ContainerCommitOptions) (types.ContainerCommitResponse, error)
```
vdemeester: Hum, there is `RepositoryName` and `Tag` in the `types.ContainerCommitOptions` struct.. It could also be replaced by a `ref`, I'll look into it.

types/reference/image_reference.go
```diff
@@ -0,0 +1,36 @@
...
+// and tags.
+func GetRepositoryAndTagFromNamedRef(ref distreference.Named) (string, string) {
+	var tag string
+	if tagged, isTagged := ref.(distreference.NamedTagged); isTagged {
+		tag = tagged.Tag()
+	}
+	// This is in order to support a reference like :
+	// name@sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
+	if digested, isDigested := ref.(distreference.Digested); isDigested {
+		tag = fmt.Sprintf("%s@%s", tag, digested.Digest().String())
```
aaronlehmann: I don't think the Docker API currently supports providing both a tag and a digest. Do you know of any places where this is allowed?

BTW, if we keep this, I would drop the `Sprintf` and do `tag += "@" + digested.Digest().String()`.

client/image_build.go
```diff
@@ -20,7 +21,7 @@ var headerRegexp = regexp.MustCompile(`\ADocker/.+\s\((.+)\)\z`)
 // ImageBuild sends request to the daemon to build images.
 // The Body in the response implement an io.ReadCloser and it's up to the caller to
 // close it.
-func (cli *Client) ImageBuild(ctx context.Context, options types.ImageBuildOptions) (types.ImageBuildResponse, error) {
+func (cli *Client) ImageBuild(ctx context.Context, context io.Reader, options types.ImageBuildOptions) (types.ImageBuildResponse, error) {
```
aaronlehmann: Having arguments named `ctx` and `context` could be confusing. Can we make `context` more specific?

vdemeester: `buildContext` maybe ? :sweat_smile:

aaronlehmann: I guess that works.

types/reference/image_reference.go
dmcgowan: I would prefer not to introduce another string format which is a subset of the reference format. If we want to support both tag and digest in the Docker API we should send it as a full reference and not try and to sneak it in with the tag.

types/reference/image_reference.go
```diff
@@ -0,0 +1,36 @@
...
+	}
+
+	repository, tag := GetRepositoryAndTagFromNamedRef(distributionRef)
+	return repository, tag, nil
+}
+
+// GetRepositoryAndTagFromNamedRef returns a repository and tag given the specified reference.
+// This function is necessary as long as the docker "server" api make the distinction between repository
+// and tags.
+func GetRepositoryAndTagFromNamedRef(ref distreference.Named) (string, string) {
```
dmcgowan: I don't see the purpose of having this return both values, this could just be `GetTagFromNamedRef`, the `Repository` portion should just be got directly from `ref.Name()`.

vdemeester: true :+1:

types/reference/image_reference.go
stevvooe: The problem here is the variable naming. The name `tag` should be `name`. The code may be wrong, as well.

dmcgowan: I think you are right, there is a disconnect between what the comment says it is supporting `name@sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff` which would assume a valid reference such as `docker.io/library/ubuntu@name@sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff` and what the code is doing `latest@name@sha256:ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff` which is not a valid reference.

vdemeester: oh that's right, I completely messed up this part, working on it :angel:

vdemeester:
@dmcgowan @aaronlehmann @stevvooe updated :angel:

stevvooe:
LGTM

I think we should put a note there that we intend to fix the fact that digest is transmitted in tag. That was not the intention. We should probably have an issue somewhere to simply take a ref over the API.

vdemeester:
@stevvooe You mean for https://github.com/docker/engine-api/pull/162/files#diff-dbeb877b7334f75bc4c9063000523d71R19 or on a more general *scale* ? (every part of the `Image*` api that has `repository` and `tag` ?)

Once this (and the `docker/docker` counter part) is merged, I'll create an issue about it and start to work this out in both repository :wink:

stevvooe:
@vdemeester I mean on a more general scale, but I am not sure about the level of complexity. The client-side parsing introduces a lot of errors but the reduced complexity may make the API easier to use in the future.

calavera:
LGTM

client/container_logs.go
```diff
@@ -13,7 +13,7 @@ import (
 
 // ContainerLogs returns the logs generated by a container in an io.ReadCloser.
 // It's up to the caller to close the stream.
-func (cli *Client) ContainerLogs(ctx context.Context, options types.ContainerLogsOptions) (io.ReadCloser, error) {
+func (cli *Client) ContainerLogs(ctx context.Context, container string, options types.ContainerLogsOptions) (io.ReadCloser, error) {
```
dongluochen: I suggest updating `container` to `containerID`? There are other places with this problem. No ambiguity is better.

dongluochen:
LGTM. I have a small change quest on parameter names.

client/container_logs.go
stevvooe: These can take a name (I think), so `containerID` is inaccurate.

vdemeester: @dongluochen I used `container` on purpose as it can be container ID or name (also did the change elsewhere).

calavera: agreed, `container` is a better name.

PR #166, opened 2016-03-21
Fix Windows storage qos hostconfig names

darrenstahlmsft:
These host config parameters are not currently used. This PR names them more correctly, as they are not exactly Block IO.

It also removes the SandboxSize option which was unused, as it fits better in the recently added StorageOpts[].


calavera:
LGTM

vdemeester:
LGTM 🐰

PR #168, opened 2016-03-23
Swagger spec for Docker APIs

sflxn:
Added a swagger spec for the Docker APIs.  This spec originated
from php-docker, massage by go-swagger, and fixed by vmware/vic.
Also added targets in the makefile to generate the swagger
client/server, build the client, and also to clean all the
generated code that gets recreated on regeneration cycles.  The
build is based on go-swagger.  The spec was tested against
go-swagger and the main swagger generator.

Signed-off-by: Loc Nguyen <nloc@vmware.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,10 @@
 
 all: deps test validate
 
+SWAGGER ?= $(GOPATH)/bin/swagger$(BIN_ARCH)
+
+swagger: $(SWAGGER)
+
 deps:
 	go get -t ./...
 	go get github.com/golang/lint/golint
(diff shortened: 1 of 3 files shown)


caglar10ur:
This is related with #77 (so that github can link them)

MHBauer:
Wow, a makefile entry and everything! Cool.

calavera:
Nice! What do we have to do to keep the spec up to date? Is there any way we can automate it?

sflxn:
Just made some changes to the README.md and Makefile to fix the checks.  There is a generated file when you generate a server that shouldn't be deleted.  I changed the Makefile so the clean target just deletes all generated files for now.

sflxn:
Whoops, last commit didn't have my changes from my last comments.  Weird.  Should be fixed now.

sflxn:
@calavera To keep it updated, Docker or the community needs to manually updated for each remote API version update.  It's an intermediate solution.  Long term, you need to chose whether you want the server code to be the source of truth or the swagger spec is the source of truth.  If you chose one path, you can always switch to the other path down the road.

If you chose the server code as the source of truth (which is the safest route for the short term), you will want to annotate the code.  Leave this spec as it is for those who want to use it to create an API client.  Use the hand made spec as a guide check for the annotation.  Annotating the code will take time to do, probably a lot more time than writing the spec by hand.  At a future time, if you feel the spec that's created from the annotation (swagger CLI call from your makefile) is at parity with the server functionality, you can then swap out the server's REST front end for one generated by Swagger... if it's a path you want to chose.

If your server code was already in this repo, I'm sure the community would be willing to assist in the annotation.

casualjim:
if you want to see what annotations you'd have to provide to annotate your current code.
you can always generate a server from the submitted specification because it will have all the doc comments already provided.
That should get you an example of what you need to put on there.

once the code is annotated with doc comments you can run `swagger generate spec` and get a new json document produced based on the code.
Currently it wants a single start package from which it can discover all the things in use (it uses golangs loader/builder infrastructure)

Feel free to submit issue or PR's at https://github.com/go-swagger/go-swagger

AndrewGuenther:
What's the status on this PR? This would be a hugely helpful change, I'd hate to see it fall through.

crosbymichael:
Thanks for this PR.

PR #169, opened 2016-03-23
Make sure the request path is properly encoded.

calavera:
Some endpoints can have special characters in the name, this ensures that we escape them properly.
See https://github.com/docker/docker/pull/21421.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,5 @@
     email: false
   go:
     - 1.6
-    - tip
   install: make deps
   script: make validate && make test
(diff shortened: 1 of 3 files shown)


vdemeester:
LGTM 🐯

MHBauer:
NEAT!

LGTM.

btw, vet issue for tip.

calavera:
I really don't get those TIP issues, this is the offending line: `tr := new(http.Transport)`

calavera:
I disabled TIP, sorry but I don't really care about vet changes in development.

MHBauer:
Allow tip to fail and not fail the build.
```
matrix:
  allow_failures:
    - go: tip
```
I think that should work, if reading the travis docs correctly.

calavera:
let's worry about Go 1.7 when it's released and we build with that version

PR #173, opened 2016-03-28
Fix for Swarm issue #2027

echupriyanov:
`NewEngine()` call contains pointer to `http.Client` structure, which contains
pointer to `http.Transport`, which in turn contains `TLSClientConfig *tls.Config`

That way, its very possible that the same TLS config can be used to communicate
with different Docker engines. And explicitly setting `ServerName` really hurts.

And we can see such situation with current master Swarm build (https://github.com/docker/swarm/issues/2027)

Signed-off-by: Eugene Chupriyanov <e.chuprianov@cpm.ru>

### DIFF
--- a/client/transport/transport.go
+++ b/client/transport/transport.go
@@ -35,10 +35,10 @@ func NewTransportWithHTTP(proto, addr string, client *http.Client) (Client, erro
 		}
 	}
 
-	if transport.TLSClientConfig != nil && transport.TLSClientConfig.ServerName == "" {
-		transport.TLSClientConfig.ServerName = hostname(addr)
-	}
-
+	// if transport.TLSClientConfig != nil && transport.TLSClientConfig.ServerName == "" {
+	// 	transport.TLSClientConfig.ServerName = hostname(addr)
+	// }
+	//
 	return &apiTransport{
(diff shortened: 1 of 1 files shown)


calavera:
This is not a proper fix, sorry. That code is valid to work with the engine.

echupriyanov:
@calavera  Can you, please, give some more details? Is it essential to explicitly set `ServerName` in caller's `TLSClientConfig`. That means Docker Swarm needs to have separate TLS configs for each engine.

calavera:
client hijacking needs the ServerName to validate the connection, afaik. But that could also be there for historic reasons, that code is very very old.

https://github.com/docker/engine-api/commit/9eec4dc67b13ace4c3c34496089122b9f1f7c28e

Removing the code, without making sure the old hijacking behavior still works as expected is not the proper solution.

echupriyanov:
@calavera Aha, I see now. What do I need to do to make sure hijacking behavior still works?

calavera:
I'm reverting my commit in #177 to fix this issue for now. That way we preserve the original behavior until we double check that the server name is necessary, and swarm works again with only one configuration.

PR #174, opened 2016-03-29
Allowing slice container filters for older daemons

### LINKED ISSUE #171: Backward compatibility for container filters
Reference: https://github.com/docker/engine-api/blob/master/types/filters/parse.go
Currently, the `Args` struct looks like this
```
type Args struct {
	fields map[string]map[string]bool
}
```
but not too long ago, it looked like this
```
type Args map[string][]string
```
So basically I can’t use the newer client to work with older daemons, because I will get a json unmarshaling error. This is a problem for Swarm since we're moving to the `engine-api` and can't support older daemons on the cluster as a result.

Current Swarm PR blocked due to this issue: https://github.com/docker/swarm/pull/1983

This issue is in the same category as #144.

nishanttotla:
This PR adds support for passing container filters to daemons older than 1.10

cc @calavera @dongluochen 

Fixes #171 

### DIFF
--- a/client/container_list.go
+++ b/client/container_list.go
@@ -35,7 +35,8 @@ func (cli *Client) ContainerList(ctx context.Context, options types.ContainerLis
 	}
 
 	if options.Filter.Len() > 0 {
-		filterJSON, err := filters.ToParam(options.Filter)
+		filterJSON, err := filters.ToParamWithVersion(cli.version, options.Filter)
+
 		if err != nil {
 			return nil, err
 		}
--- a/types/filters/parse.go
+++ b/types/filters/parse.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 )
 
@@ -62,6 +63,26 @@ func ToParam(a Args) (string, error) {
 	}
 
 	buf, err := json.Marshal(a.fields)
+	if err != nil {
+		return "", err
+	}
+	return string(buf), nil
+}
+
+func ToParamWithVersion(version string, a Args) (string, error) {
+	// this way we don't URL encode {}, just empty space
+	if a.Len() == 0 {
+		return "", nil
+	}
+
+	// for daemons older than v1.10, filter must be of the form map[string][]string
+	buf := []byte{}
+	err := errors.New("")
+	if version != "" && compareTo(version, "1.22") == -1 {
+		buf, err = json.Marshal(convertArgsToSlice(a.fields))
+	} else {
+		buf, err = json.Marshal(a.fields)
+	}
 	if err != nil {
 		return "", err
 	}
@@ -254,4 +275,49 @@ func deprecatedArgs(d map[string][]string) map[string]map[string]bool {
 		m[k] = values
 	}
 	return m
+}
+
+func convertArgsToSlice(f map[string]map[string]bool) map[string][]string {
+	m := map[string][]string{}
+	for k, v := range f {
+		values := []string{}
+		for kk, _ := range v {
+			if v[kk] {
+				values = append(values, kk)
+			}
+		}
+		m[k] = values
+	}
+	return m
+}
+
+// compareTo compares two version strings
+// returns -1 if v1 < v2, 1 if v1 > v2, 0 otherwise
+func compareTo(v1, v2 string) int {
+	var (
+		currTab  = strings.Split(v1, ".")
+		otherTab = strings.Split(v2, ".")
+	)
+
+	max := len(currTab)
+	if len(otherTab) > max {
+		max = len(otherTab)
+	}
+	for i := 0; i < max; i++ {
+		var currInt, otherInt int
+
+		if len(currTab) > i {
+			currInt, _ = strconv.Atoi(currTab[i])
+		}
+		if len(otherTab) > i {
+			otherInt, _ = strconv.Atoi(otherTab[i])
+		}
+		if currInt > otherInt {
+			return 1
+		}
+		if otherInt > currInt {
+			return -1
+		}
+	}
+	return 0
 }
--- a/types/filters/parse_test.go
+++ b/types/filters/parse_test.go
@@ -57,6 +57,31 @@ func TestToParam(t *testing.T) {
 	}
 }
 
+func TestToParamWithVersion(t *testing.T) {
+	fields := map[string]map[string]bool{
+		"created":    {"today": true},
+		"image.name": {"ubuntu*": true, "*untu": true},
+	}
+	a := Args{fields: fields}
+
+	str1, err := ToParamWithVersion("1.21", a)
+	if err != nil {
+		t.Errorf("failed to marshal the filters with version < 1.22: %s", err)
+	}
+	str2, err := ToParamWithVersion("1.22", a)
+	if err != nil {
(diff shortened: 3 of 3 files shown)


calavera:
I'd rather not use the word "deprecated", we can use the version in the client to decide how to serialize the arguments, as I mentioned in the original issue.

nishanttotla:
@calavera can you take a look now? I added a check in the client to call the appropriate serialization function.

One issue that still remains is that with Swarm, the server version is only found out after the client has been created. Is there a way to update the version later? If not, could we add a function to do that?

client/container_list.go
```diff
@@ -35,7 +37,14 @@ func (cli *Client) ContainerList(ctx context.Context, options types.ContainerLis
 	}
 
 	if options.Filter.Len() > 0 {
-		filterJSON, err := filters.ToParam(options.Filter)
+		filterJSON := ""
+		err := errors.New("")
+		// for daemons older than v1.10, filter must be of the form map[string][]string
+		if cli.version != "" && version.Version(cli.version).LessThan(version.Version("1.10")) {
```
dongluochen: `1.10` -> CLI should deal with API version, not Engine version. 
https://docs.docker.com/engine/reference/api/docker_remote_api/

calavera: I'd also rather not have this conditional here. What about calling `filters.ToParamWithVersion(cli.version, options.Filters)` so all this logic is encapsulated there?

client/container_list.go
```diff
@@ -2,9 +2,11 @@ package client
 
 import (
 	"encoding/json"
+	"errors"
 	"net/url"
 	"strconv"
 
+	"github.com/docker/docker/pkg/version"
```
calavera: we cannot import a docker package, it's design constraint, sorry. We should probably extract that package from docker or use a different one.

nishanttotla: Okay I understand, since `engine-api` is vendored into Docker? I'll try to find another package, but I suppose long term it would be best to use the same one.

client/container_list.go
nishanttotla: @dongluochen The problem I'm facing is that one client (with a constant API version) will make a call to the Swarm master, which will then make a call via `engine-api` to a daemon that could be old or new. In such a situation, how does Swarm indicate to the client what API version to use?

@calavera Not an issue. I didn't do that right now because `ToParam` is exported, so I wasn't sure we wanted to change it. WDYT?

calavera: > I wasn't sure we wanted to change it

I didn't say to change it, I'd prefer to add a new function to the filters package that handle it.

> how does Swarm indicate to the client what API version to use?

tbh, the fact that we're pushing this to the client sounds like a bad idea. Swarm should handle any transformation it needs to do to call to old daemons. It looks like you're trying to use the same client to talk with all daemons, I'm not sure it's the best approach.

nishanttotla: Okay, new function makes sense.

I agree that it's a bad idea in general. In Swarm, we can certainly handle the transformation. There's just one thing that is missing then - the `version` in the apiClient needs to be updated from time to time depending on which daemon is being talked to (Swarm will handle this). But is that currently doable in `engine-api`?

calavera: why don't you have a pool of clients? you're trying to manage a pool of daemon versions with a single apiClient, is that a requirement? the httpClient can still be reused between apiClients.

nishanttotla: Sorry I think I wasn't very clear before.

For each engine in the cluster, there indeed is a dedicated apiClient for that engine, but with two extra details:
1. We only find out the engine version after the client has been created
2. Since engines in the cluster can go down and come back up after being updated, we'd like to be able to update the API version if needed.

Beyond these two cases, things are fine, and work exactly as you pointed out.

client/container_list.go
nishanttotla: @calavera regarding the version package, what is your opinion about creating the same one inside of `engine-api` to get rid of the dependency on `docker/docker`? It's a relatively small package, and I feel it's unnecessary to add external dependency just for that.

client/container_list.go
calavera: I see, it makes more sense now.

types/filters/parse.go
```diff
@@ -68,6 +70,26 @@ func ToParam(a Args) (string, error) {
...
+func ToParamWithVersion(version string, a Args) (string, error) {
+	// this way we don't URL encode {}, just empty space
+	if a.Len() == 0 {
+		return "", nil
+	}
+
+	// for daemons older than v1.10, filter must be of the form map[string][]string
+	buf := []byte{}
+	err := errors.New("")
+	if version != "" && versionpkg.Version(version).LessThan(versionpkg.Version("1.10")) {
```
calavera: this won't ever be true, because `cli.version` is the API version.

types/filters/parse.go
```diff
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+
+	versionpkg "github.com/docker/docker/pkg/version"
```
calavera: as I mentioned, we cannot import from `docker/docker`. It's a restriction of this package.

calavera: you can just copy this function and use it to compare the two strings directly:

https://github.com/docker/docker/blob/master/pkg/version/version.go#L11-L38

nishanttotla: Thanks, that was my intention too, fixing it now. I mentioned it in an earlier comment, maybe it got suppressed.

types/filters/parse.go
nishanttotla: My bad, I forgot to update this part. Also, is it acceptable to add a version update function in the client?

calavera: yeah, it should be fine

nishanttotla:
@calavera I've fixed issues based on our discussion, and opened #185 for the client version update function.

types/filters/parse.go
```diff
@@ -68,6 +69,26 @@ func ToParam(a Args) (string, error) {
...
+func ToParamWithVersion(version string, a Args) (string, error) {
+	// this way we don't URL encode {}, just empty space
+	if a.Len() == 0 {
+		return "", nil
+	}
+
+	// for daemons older than v1.10, filter must be of the form map[string][]string
+	buf := []byte{}
+	err := errors.New("")
+	if version != "" && compareTo(version, "1.22") == -1 {
```
dongluochen: I think you mean "1.22 and above". It should be version `LessThan 1.22`, otherwise `1.23` will be wrong.

types/filters/parse.go
```diff
@@ -255,3 +276,48 @@ func deprecatedArgs(d map[string][]string) map[string]map[string]bool {
...
+			if v[kk] {
+				values = append(values, kk)
+			}
+		}
+		m[k] = values
+	}
+	return m
+}
+
+// CompareTo compares two version strings
```
dongluochen: nit: `CompareTo` -> `compareTo`

nishanttotla: fixed, thanks!

dongluochen:
LGTM

nishanttotla:
Ping @calavera @vdemeester

vdemeester:
LGTM 🐵

nishanttotla:
Could we merge this?

PR #175, opened 2016-03-29
Add checkpoint/restore interface to the engine API.

boucher:
An initial attempt at introducing an interface for checkpoint/restore to the engine API. Currently, the only checkpoint option is "exit", and the only restore option is the checkpoint id to restore. I'd love to hear from @crosbymichael about the containerd level options and what we'd like to see in Docker.

Containerd currently doesn't seem to support a specific path for checkpoint images, which would be really helpful for my own use case. The other options currently in containerd seem potentially ok to leave out.

### DIFF
--- a/client/container_checkpoint.go
+++ b/client/container_checkpoint.go
@@ -9,15 +9,16 @@ import (
 )
 
 // ContainerCheckpoint checkpoints a running container
-func (cli *Client) ContainerCheckpoint(ctx context.Context, containerID string, exit bool) (types.ContainerCheckpointResponse, error) {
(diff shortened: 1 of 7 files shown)


LK4D4:
@boucher thanks for a patch. I think we should go with config structs right now; there will be probably much more configuration for c/r. Am I right?

boucher:
Sounds reasonable to me. I'll make the change once I get the chance. I don't really know what if any plans @crosbymichael has for containerd, so I am interested in his feedback. At the very least I actually think a user specified name for the checkpoint is really valuable.

boucher:
Failure seems to be from master.

vdemeester:
I guess this one should be closed in favor of #202, if not please ping me to reopen it 😇

PR #177, opened 2016-03-30
Revert "Set the TLS connection ServerName when the transport is created."

calavera:
This commit is causing issues in Swarm. We don't need this change yet, so I rather revert it for the upcoming release.

This reverts commit 9eec4dc67b13ace4c3c34496089122b9f1f7c28e.


vdemeester:
Arf :sweat: 
LGTM 🐡

LK4D4:
LGTM

PR #183, opened 2016-03-31
Throw client.Do(req) error in postHijacked

WeiZhang555:
Related docker PR: https://github.com/docker/docker/pull/21092

Sometime when client issues an http request via clientconn.Do(req),
server can send an error through hijacked http connection and close this
connection afterwards. We need to throw this error(httputil.ErrPersistEOF)
out, to let client know the connection is closed, so client can choose
to abort after reading detailed error message from hijacked connection.

Also client can ignore other kinds of errors, so it won't bring any problem.
And it will make more sense to not swallow the error.

### DIFF
--- a/client/hijack.go
+++ b/client/hijack.go
@@ -68,11 +68,11 @@ func (cli *Client) postHijacked(ctx context.Context, path string, query url.Valu
 	defer clientconn.Close()
 
 	// Server hijacks the connection, error 'connection closed' expected
-	clientconn.Do(req)
+	_, err = clientconn.Do(req)
 
 	rwc, br := clientconn.Hijack()
 
-	return types.HijackedResponse{Conn: rwc, Reader: br}, nil
+	return types.HijackedResponse{Conn: rwc, Reader: br}, err
 }
 
 func tlsDial(network, addr string, config *tls.Config) (net.Conn, error) {


vdemeester:
LGTM 🐙
/cc @calavera

client/hijack.go
```diff
@@ -68,11 +68,11 @@ func (cli *Client) postHijacked(ctx context.Context, path string, query url.Valu
 	defer clientconn.Close()
 
 	// Server hijacks the connection, error 'connection closed' expected
-	clientconn.Do(req)
+	_, err = clientconn.Do(req)
```
calavera: should we return here before calling `Hijack()` if `err != nil`?

vdemeester: ah that's a good point I think.

WeiZhang555: No, hijack is still necessary. The case is, server meets an error, it sends error message via Hijacked connection then close it. Client gets an ErrPersistError, but it can still read error message from low level tcp connection.
So we need to do Hijack() even though error happened, only need to throw it out and let client choose how to handle the error

calavera:
LGTM

client/hijack.go
vdemeester: ah make sence

PR #185, opened 2016-04-01
Adding function to allow for updating client version

nishanttotla:
This is related to the discussion in #174 

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -1,8 +1,16 @@
 package client
 
 import (
+	"bytes"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
 	"net/url"
+	"strings"
 	"testing"
+
+	"github.com/docker/engine-api/types"
+	"golang.org/x/net/context"
 )
 
 func TestGetAPIPath(t *testing.T) {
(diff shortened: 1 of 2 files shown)


dongluochen:
LGTM

duglin:
I know its a trivial change/function - but can we get a testcase for this?

nishanttotla:
@duglin did you have something particular in mind? There's no logic in the function, so I'm not sure what we'd test.

duglin:
I was thinking of making sure that the version string appears correctly in the outgoing http request.

nishanttotla:
@duglin Just added tests. Please take a look :)

client/client_test.go
```diff
@@ -70,3 +70,30 @@ func TestParseHost(t *testing.T) {
...
+	}
+
+	apiClient, err := NewClient("unix:///var/run/docker.sock", "", nil, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, cs := range cases {
+		apiClient.UpdateClientVersion(cs.v)
+		g := apiClient.getAPIPath(cs.p, cs.q)
```
duglin: My concern with checking getAPIPath instead of actually checking what the server sees is that if someone adds a new code path to open a connection then we'll miss it.  Might not be a big deal but I'd like to hear what @vdemeester thinks.
See https://github.com/docker/engine-api/blob/master/client/container_create_test.go for an example of checking it from the server side.

client/client_test.go
```diff
@@ -70,3 +70,30 @@ func TestParseHost(t *testing.T) {
...
+
+func TestUpdateClientVersion(t *testing.T) {
+	cases := []struct {
+		v string
+		p string
+		q url.Values
+		e string
+	}{
+		{"1.20", "/containers/json", nil, "/v1.20/containers/json"},
+		{"v1.21", "/containers/json", url.Values{}, "/v1.21/containers/json"},
```
duglin: Interesting that w/ and w/o a leading "v" will generate the same result - not sure I would have done that, but that's not part of this PR  :-)

nishanttotla: Yeah, I just looked at the test above, and assumed it was required/expected behavior.

client/client_test.go
vdemeester: I do agree with @duglin, I think it would better to check *server side* :angel:

nishanttotla:
@duglin @vdemeester updated, please let me know if this works!

client/client_test.go
```diff
@@ -70,3 +78,43 @@ func TestParseHost(t *testing.T) {
 		}
 	}
 }
+
+func TestUpdateClientVersion(t *testing.T) {
+	client := &Client{
+		transport: newMockClient(nil, func(req *http.Request) (*http.Response, error) {
+			reqString := req.URL.String()
+			splitQuery := strings.Split(reqString, "/")
+			queryVersion := splitQuery[3]
```
duglin: I think you'd be better of using req.URL.Path instead of req.URL.String() then you don't need to worry about any leading /'s that are before the `/vX.Y` part of the URL.  And then you know that the version part is always the first thing in that string.

duglin:
One minor request on the server-side parsing of the incoming URL.
Aside from that LGTM

vdemeester:
Same nit as @duglin, otherwise LGTM 🐰

nishanttotla:
@duglin @vdemeester that makes sense, thanks. It's updated now.

duglin:
still LGTM
ping @vdemeester for the re-LGTM and merge

vdemeester:
ah yes, re-LGTM :angel:

PR #186, opened 2016-04-01
Don't claim to have json payload when it is blank

bboreham:
https://github.com/docker/engine-api/commit/002fcc738c1ef307200e0ac1cb86a3306d9ffe8a#diff-ac3eb5494dc3178ea5947441c086942bR59 had the unfortunate effect of changing the behaviour of the second if-statement: it is meant to skip setting the Content-Type if the body is actually empty.

An empty string is not valid JSON so some recipients of the data may object, e.g. as found in https://github.com/weaveworks/weave/issues/2109

### DIFF
--- a/client/container_start_test.go
+++ b/client/container_start_test.go
@@ -2,6 +2,8 @@ package client
 
 import (
 	"bytes"
+	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"testing"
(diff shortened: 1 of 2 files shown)


MHBauer:
LGTM in the sense that this will change the functionality back to what it was in the referenced commit.

I feel the code looks a little funky and could maybe be reordered, but I'm a little hesitant to make a suggestion. Is there a unit test we can write to check this "don't set the `Content-Type` if the body is empty"?

anusha-ragunathan:
LGTM

bboreham:
I have added a test, based off the scenario where we detected this issue.  We could have a lower-level test more tightly bound to the specific error, but since there isn't a `request_test.go` it would be a rather larger exercise to create one worthy of the name.

Per @MHBauer suggestion I also offer a re-ordering of `sendRequest()` which is slightly more efficient and makes the intent clearer in the `nil` case.

MHBauer:
LGTM. 

I like seeing the commits.

tiborvass:
LGTM

Issue #189, opened 2016-04-04
Engine-api built with Go(<=1.4) won't work with Docker daemon built with Go(>=1.6)


Random-Liu:
When using engine-api built with Go(<=1.4) with Docker daemon built with Go(>=1.6), it keeps complaining `"Error response from daemon: 400 Bad Request: malformed Host header"`.

I think the reason is that:

1. After https://github.com/golang/go/commit/6e11f45ebdbc7b0ee1367c80ea0a0c0ec52d6db5, the http server in Go 1.6 will validate the [HTTP Host header](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23) conforming to HTTP 1.1 (See https://github.com/golang/go/issues/13624)
2. On the http client side, Go writes the HTTP Host header with `request.Host` or `request.URL.Host`(if the former is empty) (See [here](https://github.com/golang/go/blob/a6557a05a03490af3b26f97f9a4ce99c7c773fe5/src/net/http/request.go#L404-L410))
3. When using the [default docker endpoint](https://github.com/docker/engine-api/blob/68a7b6bebf8f57d559b7788a46c55045438747b9/client/client_unix.go#L6) (`unix:///var/run/docker.sock`), engine-api will set `request.Host=""`, `request.URL.Host="/var/run/docker.sock"` (See [set request.URL.Host](https://github.com/docker/engine-api/blob/68a7b6bebf8f57d559b7788a46c55045438747b9/client/request.go#L85-L87) and [set request.Host](https://github.com/docker/engine-api/blob/68a7b6bebf8f57d559b7788a46c55045438747b9/client/request.go#L131))
4. When the engine-api is built with Go >= 1.5, everything goes well, because Go will [clean up the Host](https://github.com/golang/go/blob/master/src/net/http/request.go#L523) making both `request.Host` and `request.URL.Host` to "". Then the HTTP Host header will be "". (See https://github.com/golang/go/issues/11206))
5. But when the engine-api is built with Go<=1.4, Go [won't clean up the Host](https://github.com/golang/go/blob/release-branch.go1.4/src/net/http/request.go#L368-L373), so the HTTP Host header will be `"/var/run/docker.sock"`. It will break Go 1.6 http server side validation in 1. because `'/'`is not a valid character, causing `400 Bad Request: malformed Host header`.

[...]

PR #190, opened 2016-04-04
Set fake host header for local communication.

Random-Liu:
For issue #189.

In this PR, I set `URL.Host` to a fake address "docker.sock" to solve #189.

Hope this could be in, thanks! :)

@duglin @vdemeester

### DIFF
--- a/client/request.go
+++ b/client/request.go
@@ -83,6 +83,11 @@ func (cli *Client) sendClientRequest(ctx context.Context, method, path string, q
 	}
 
 	req, err := cli.newRequest(method, path, query, body, headers)
+	if cli.proto == "unix" || cli.proto == "npipe" {
+		// For local communications, it doesn't matter what the host is. We just
+		// need a valid and meaningful host name. (See #189)
+		req.Host = "docker"
+	}
 	req.URL.Host = cli.addr
 	req.URL.Scheme = cli.transport.Scheme()
 
--- a/client/request_test.go
+++ b/client/request_test.go
@@ -1,0 +1,77 @@
+package client
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"testing"
+
+	"golang.org/x/net/context"
+)
+
+// TestSetHostHeader should set fake host for local communications, set real host
+// for normal communications.
+func TestSetHostHeader(t *testing.T) {
+	testURL := "/test"
+	testCases := []struct {
+		host            string
+		expectedHost    string
+		expectedURLHost string
+	}{
+		{
+			"unix:///var/run/docker.sock",
+			"docker",
+			"/var/run/docker.sock",
+		},
+		{
+			"npipe:////./pipe/docker_engine",
+			"docker",
+			"//./pipe/docker_engine",
+		},
+		{
+			"tcp://0.0.0.0:4243",
+			"",
+			"0.0.0.0:4243",
+		},
+		{
+			"tcp://localhost:4243",
+			"",
+			"localhost:4243",
+		},
+	}
+
+	for c, test := range testCases {
+		proto, addr, basePath, err := ParseHost(test.host)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		client := &Client{
+			transport: newMockClient(nil, func(req *http.Request) (*http.Response, error) {
+				if !strings.HasPrefix(req.URL.Path, testURL) {
+					return nil, fmt.Errorf("Test Case #%d: Expected URL %q, got %q", c, testURL, req.URL)
+				}
+				if req.Host != test.expectedHost {
+					return nil, fmt.Errorf("Test Case #%d: Expected host %q, got %q", c, test.expectedHost, req.Host)
+				}
+				if req.URL.Host != test.expectedURLHost {
+					return nil, fmt.Errorf("Test Case #%d: Expected URL host %q, got %q", c, test.expectedURLHost, req.URL.Host)
+				}
+				return &http.Response{
+					StatusCode: http.StatusOK,
+					Body:       ioutil.NopCloser(bytes.NewReader(([]byte("")))),
+				}, nil
+			}),
+			proto:    proto,
+			addr:     addr,
+			basePath: basePath,
+		}
+
+		_, err = client.sendRequest(context.Background(), "GET", testURL, nil, nil, nil)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+}


Random-Liu:
There is something wrong for windows support. Will update this PR soon.

duglin:
Would it be possible to add a testcase for this to make sure we don't change it (or remove it) by accident?
Aside from that it looks good.

Random-Liu:
@duglin Sure, will do that soon. Thanks!

Random-Liu:
@duglin Addressed comments.

client/request_test.go
```diff
@@ -0,0 +1,72 @@
...
+			"docker",
+		},
+		{
+			"npipe:////./pipe/docker_engine",
+			"",
+			"docker",
+		},
+		{
+			"tcp://0.0.0.0:4243",
+			"",
```
duglin: Even though the PR is focused on the `""` case, let's make the testcase cover the "normal" host cases too. Can you please add a test variant where `test.expectedHost` isn't `""`?

Random-Liu: I think the `test.expectedHost` is always `""`.
Now we just pass the apiPath (something with no schema like `/container/id/start`) into `http.NewRequest`. (See [here](https://github.com/docker/engine-api/blob/master/client/request.go#L132-L133))
While `http.NewRequest` will only set `request.Host` when the passed in url has Schema. (See [here](https://github.com/golang/go/blob/c1c7547f6ad7264c1d6eea3fc1645b2eab104694/src/net/http/request.go#L651), [here](https://github.com/golang/go/blob/c1c7547f6ad7264c1d6eea3fc1645b2eab104694/src/net/url/url.go#L417) and [here](https://github.com/golang/go/blob/c1c7547f6ad7264c1d6eea3fc1645b2eab104694/src/net/url/url.go#L483-L490).

Random-Liu: Oh, wait, let me see.

duglin: wouldn't using "tcp://localhost:4243" do it?

Random-Liu: No. The `request.Host` is still `""`.

Random-Liu: OK. If we are using something like "tcp://0.0.0.0:4243//abc/efg", the `request.Host` will be set to `"abc"`.

Because the `basePath` will become "//abc/efg", then the `apiPath` will become something like "//abc/efg/container/id/start", then `http.NewRequest` will set `request.Host` to "abc".

I think this is also a bug in engine-api, we should fix it instead of testing it.

duglin: I think the issue is with the mock client not the engine-api code itself. The mock client doesn't go thru to same codepath to set the http info.  I modified the test and I think I now see the correct results.
See: https://gist.github.com/duglin/4596113b600bcf44c84edd00b2368aa2
In particular, line 46 followed by setting the transport.

ping @vdemeester to get his opinion and if this is "more correct" then we should probably modify the other testcases to use (aka test) `NewClient()`

duglin: oh I also added:
```
req.Host = host
```
around like 92 in request.go

vdemeester: oh yep, I think it would be better to use `NewClient` everywhere.

- renaming `newMockClient` to `newMockTransportClient` (or `newMockTransport`…) as it's more what it does.
- probably creating a `newMockClient` that does call `NewClient`, …

```go
func newMockClient(tlsConfig *tls.Config, doer func(*http.Request) (*http.Response, error)) Client {
        client, _ := NewClient(test.host, "", nil, nil)

        client.transport = newMockTransport(tlsConfig, doer)

        return client
}
```

I would even think we should check the error on `NewClient`.

Random-Liu: @duglin, I think the main reason is that you added `req.Host = host` in request.go, there is nothing to do with the `NewClient`.

The `req.Host` is initialized in `http.NewRequest` before, and mostly it will be set to "" except the bug I mentioned [above](https://github.com/docker/engine-api/pull/190/files#r58492191).

Now you manually added `req.Host = host`, it will be sure to be the same with `req.URL.Host`.

Random-Liu:
@duglin @vdemeester Ping~ :P

vdemeester:
LGTM 🐶

client/request_test.go
duglin: Something feels wrong here.  According to this test the HTTP `Host` header will always be empty, which doesn't seem right to me.  If it really is supposed to be blank then why include the string in the test data, just check for a static `""` in the if-statement. However, having an empty header is not correct and we should fix that.

I'll try to find some time to look at this but kind of busy today....

Random-Liu: @duglin Yeah, you are right. If it is supposed to be blank, we should just check it with a static `""`. There is indeed something wrong here, wait for your result. :)

client/request.go
```diff
@@ -83,7 +83,13 @@ func (cli *Client) sendClientRequest(ctx context.Context, method, path string, q
 	}
 
 	req, err := cli.newRequest(method, path, query, body, headers)
-	req.URL.Host = cli.addr
+	host := cli.addr
+	if cli.proto == "unix" || cli.proto == "npipe" {
+		// For local communications, it doesn't matter what the host is. We just
+		// need a valid and meaningful host name. (See #189)
+		host = "docker"
```
MHBauer: shouldn't this set `req.Host`? We override the header (the thing that causes the problem on the server) only in the case that we need to talk to a Host that would cause a problem. The request still needs to go to the right place.

Random-Liu: @MHBauer Make sense to me, will update soon.

Random-Liu:
@MHBauer Addressed comments.

MHBauer:
@Random-Liu thanks for the changes. I appreciate the comments as well. If I don't make sense call me on it.

I think the three way tests were better. In my opinion the test data should look like:
```
		{
			"unix:///var/run/docker.sock",
			"docker",
			"/var/run/docker.sock",
		},
		{
			"npipe:////./pipe/docker_engine",
			"docker",
			"//./pipe/docker_engine",
		},
		{
			"tcp://0.0.0.0:4243",
			"0.0.0.0:4243",
			"0.0.0.0:4243",
		},
		{
			"tcp://localhost:4243",
			"localhost:4243",
			"localhost:4243",
		},
```
My understanding is the actual request will look different from the `Host` header in the case of sockets. We should check that.

@vdemeester @duglin Am I on to something or misinterpreting?

Random-Liu:
@MHBauer Unfortunately, `request.Host` is always `""` now. With the current test case, the result is
```
docker
/var/run/docker.sock

docker
//./pipe/docker_engine


0.0.0.0:4243
```

MHBauer:
Yea, I think that's an artifact of using the mock client. The request never actually gets to the `golang write()` method which you linked in the original issue #189. Thus the `req.URL.Host` -> `req.Host` transformation never actually happens in the testing scenario.

The change feels correct to me, but I'm not sure the test is worthwhile if it doesn't connect to an actual server and go through the whole end to end integration scenario.

MHBauer:
Understanding it further, I would expect it to be:
```
		{
			"unix:///var/run/docker.sock",
			"docker",
			"/var/run/docker.sock",
		},
		{
			"npipe:////./pipe/docker_engine",
			"docker",
			"//./pipe/docker_engine",
		},
		{
			"tcp://0.0.0.0:4243",
			"",
			"0.0.0.0:4243",
		},
		{
			"tcp://localhost:4243",
			"",
			"localhost:4243",
		},
	}
```

MHBauer:
LGTM for this as it is, based on my understanding.

Random-Liu:
@MHBauer Will address comment soon.

duglin:
I agree with @MHBauer about not using a real server for the testing. It feels like we're debugging/testing the testing code and not docker.  Did anyone consider using a real http server in the tests like we do here:  https://github.com/docker/docker/blob/master/integration-cli/docker_cli_config_test.go#L26  ?

calavera:
@duglin if the server part ended up in this repo it would be really easy to use as a test harness. We'd only need to provide test backends and we'd get end to end tests.

This change LGTM. I'm going to merge it as it is, but we can keep the discussion going in an issue.

MHBauer:
Thank you for the contribution @Random-Liu!

Random-Liu:
@MHBauer Thanks for your review and good suggestions. :)

PR #195, opened 2016-04-08
Move Docker version comparisons to the types package.

calavera:
Most of the comparisons are done at the API level.
In the long term, it makes more sense to have these functions
in this package, so people can reuse them implementing other clients.

This change removes the custom Version type. The custom type doesn't
provide any extra value since you can wrap any string in it an the
comparisons can easily be done with plain strings.

I'll proceed to make the necessary changes in docker/docker if this gets merged.

### DIFF
--- a/types/filters/parse.go
+++ b/types/filters/parse.go
@@ -7,8 +7,9 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
-	"strconv"
 	"strings"
+
+	"github.com/docker/engine-api/types/versions"
 )
 
 // Args stores filter arguments as map key:{map key: bool}.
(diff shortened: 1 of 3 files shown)


thaJeztah:
sgtm

vieux:
SGTM

types/versions/compare.go
```diff
@@ -0,0 +1,62 @@
+package versions
+
+import (
+	"strconv"
+	"strings"
+)
+
+// compare compares two version strings
+// returns -1 if v1 < v2, 1 if v1 > v2, 0 otherwise.
+func compare(v1, v2 string) int {
```
runcom: this is like the same as https://github.com/docker/docker/blob/master/pkg/version/version.go#L11

cpuguy83:
LGTM

types/filters/parse.go
```diff
@@ -80,7 +81,7 @@ func ToParamWithVersion(version string, a Args) (string, error) {
 	// for daemons older than v1.10, filter must be of the form map[string][]string
 	buf := []byte{}
 	err := errors.New("")
-	if version != "" && compareTo(version, "1.22") == -1 {
+	if version != "" && versions.LessThan(version, "1.22") {
```
runcom: can we cast version to pkg/version and do the check with pkg/version code instead?

types/versions/compare.go
runcom: or either remove pkg/version (?)

calavera: @runcom the idea is to remove the package, it's in the description.

tiborvass:
LGTM

PR #200, opened 2016-04-13
make restartpolicy.IsNone return true if restart policy name is null

coolljt0725:
if create a container via API and does not set restart polciy, the restart
name is "", it should also be "no" restart policy.

relate to https://github.com/docker/docker/pull/21993
Signed-off-by: Lei Jitang <leijitang@huawei.com>


calavera:
LGTM

vdemeester:
LGTM 🐮

PR #202, opened 2016-04-14
Add Checkpoint Create/List/Delete methods.

boucher:
Maps directly to the API exposed by containerd.
This is an alternative approach to: https://github.com/docker/engine-api/pull/175

### DIFF
--- a/client/checkpoint_create.go
+++ b/client/checkpoint_create.go
@@ -1,0 +1,13 @@
+package client
+
+import (
+	"github.com/docker/engine-api/types"
+	"golang.org/x/net/context"
+)
+
+// CheckpointCreate creates a checkpoint from the given container with the given name
+func (cli *Client) CheckpointCreate(ctx context.Context, container string, options types.CheckpointCreateOptions) error {
+	resp, err := cli.post(ctx, "/containers/"+container+"/checkpoints", nil, options, nil)
+	ensureReaderClosed(resp)
+	return err
+}
--- a/client/interface.go
+++ b/client/interface.go
@@ -15,6 +15,9 @@ import (
 // APIClient is an interface that clients that talk with a docker server must implement.
 type APIClient interface {
 	ClientVersion() string
+	CheckpointCreate(ctx context.Context, container string, options types.CheckpointCreateOptions) error
+	CheckpointDelete(ctx context.Context, container string, checkpointID string) error
+	CheckpointList(ctx context.Context, container string) ([]types.Checkpoint, error)
 	ContainerAttach(ctx context.Context, container string, options types.ContainerAttachOptions) (types.HijackedResponse, error)
 	ContainerCommit(ctx context.Context, container string, options types.ContainerCommitOptions) (types.ContainerCommitResponse, error)
 	ContainerCreate(ctx context.Context, config *container.Config, hostConfig *container.HostConfig, networkingConfig *network.NetworkingConfig, containerName string) (types.ContainerCreateResponse, error)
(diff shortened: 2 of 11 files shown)


client/checkpoint_create.go
```diff
@@ -0,0 +1,13 @@
+package client
+
+import (
+	"github.com/docker/engine-api/types"
+	"golang.org/x/net/context"
+)
+
+// CheckpointCreate creates a checkpoint from the given container with the given name
+func (cli *Client) CheckpointCreate(ctx context.Context, options types.CheckpointCreateOptions) error {
```
vdemeester: The signature should be `CheckpointCreate(ctx context.Context, container string, options types.CheckpointCreateOptions)` (and `CheckpointCreateOptions` should not have a `ContainerID` attribute). The container ID or Name is required and thus should not be considered as an option (it make it easier to consume the API too, you know it is required to provide a container Id or Name).

client/interface.go
```diff
@@ -15,8 +15,9 @@ import (
 // APIClient is an interface that clients that talk with a docker server must implement.
 type APIClient interface {
 	ClientVersion() string
-	ContainerAttach(ctx context.Context, container string, options types.ContainerAttachOptions) (types.HijackedResponse, error)
```
vdemeester: Why `ContainerAttach` and `ContainerCommit` disappear ? (typos :stuck_out_tongue_closed_eyes:)

boucher: Rebase error. Just pushed the fix.

SaiedKazemi:
@boucher @crosbymichael with docker 1.11 built on containerd already out, is there anything we can do to have the C/R pull requests reviewed?

xemul:
@crosbymichael , I +1 @SaiedKazemi 's comment :)

MHBauer:
So do we need both this and #175, or just this?

boucher:
This or #175 but not both.

LK4D4:
LGTM design

calavera:
LGTM

icecrime:
@boucher Any particular reason why restore isn't part of this PR?

boucher:
@icecrime It is -- it's treated the same way as in containerd: a checkpoint name argument to the container create call. 

I'm not very happy with that though, because it makes it a lot more difficult to actually use; you've got to copy or move the checkpoint manually.

crosbymichael:
You don't have to map the docker api directly to how containerd does it but what would be the other option that you think is better?

boucher:
@crosbymichael it seems like at the very least, if I have a container that I checkpoint I should be able to restore it without creating a new one? 

In my ideal world, though, I'd be able to just pass a path to the criu image, since I actually want to manage where I'm storing these images myself.

crosbymichael:
ya, so the checkpoint options should be passed on start, not create?

boucher:
Yeah, that seems reasonable: 

```
docker checkpoint foo checkpoint1
docker start foo --checkpoint=checkpoint1
```

Any thoughts on the path issue?

crosbymichael:
that was the original idea

crosbymichael:
I think for now, moving the restore to start would solve alot of the initial issues.  I don't know about that paths right now but if its an option we can look at adding that later

boucher:
Moved checkpoint to be an argument to container start.

types/types.go
```diff
@@ -471,3 +471,13 @@ type NetworkDisconnect struct {
...
 	Force     bool
 }
+
+// Checkpoint represents the details of a checkpoint
+type Checkpoint struct {
+	Name string // Name is the name of the checkpoint
+}
+
+// CheckpointListResponse contains the response for the remote API:
+type CheckpointListResponse struct {
```
crosbymichael: Why are we returning a type instead of `[]Checkpoint` ?

boucher: No good reason. Will change.

crosbymichael:
LGTM

calavera:
still LGTM

PR #203, opened 2016-04-15
Update CopyToContainer method signature…

vdemeester:
… to define the required arguments needed to be called 🐙. It makes it closer to `ContainerStatPath` too.

I think I missed this one from my previous PR :sweat:.

/cc @stevvooe @calavera @MHBauer 

🐸 

### DIFF
--- a/client/container_copy.go
+++ b/client/container_copy.go
@@ -30,17 +30,17 @@ func (cli *Client) ContainerStatPath(ctx context.Context, containerID, path stri
 }
 
 // CopyToContainer copies content into the container filesystem.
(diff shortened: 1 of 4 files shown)


calavera:
LGTM

vdemeester:
Hum, something's weird with appveyor these days :sweat:

calavera:
vendoring dostribution was not a very good idea, I guess 😔

stevvooe:
LGTM

cc @jlhawn

@calavera 

> vendoring distribution was not a very good idea, I guess 😔

Tsk, tsk. Don't be silly. Looks like there is a filename that is not valid in windows. Maybe that needs to be changed?

Also, why is this using `go get -t` for dependencies?

PR #205, opened 2016-04-17
Update GetTagFromNamedRef to return latest…

vdemeester:
… in case of no tag in the reference 🐰. 

Right now, if you pass a reference without a tag will send a request without tag to the daemon and thus the daemon will download all tags for this repository. `docker/docker` adds the default `latest` tag if it has not been and if `--all` is not defined.

I'm not sure this is the right place for it, waiting for your input. This change would break the way `--all` works, and the 2nd commits adds a more explicit way to specify we want to download all tags (putting it in option). If we decide to go this route, I'll do the PR in `docker/docker` to update the server side API.

/cc @stevvooe @calavera @MHBauer 

This also adds some unit test for this package 🐙.

🐸

### DIFF
--- a/types/reference/image_reference_test.go
+++ b/types/reference/image_reference_test.go
@@ -1,0 +1,72 @@
+package reference
+
+import (
+	"testing"
+)
+
+func TestParse(t *testing.T) {
+	testCases := []struct {
+		ref           string
+		expectedName  string
+		expectedTag   string
+		expectedError bool
+	}{
+		{
+			ref:           "",
+			expectedName:  "",
+			expectedTag:   "",
+			expectedError: true,
+		},
+		{
+			ref:           "repository",
+			expectedName:  "repository",
+			expectedTag:   "latest",
+			expectedError: false,
+		},
+		{
+			ref:           "repository:tag",
+			expectedName:  "repository",
+			expectedTag:   "tag",
+			expectedError: false,
+		},
+		{
+			ref:           "test.com/repository",
+			expectedName:  "test.com/repository",
+			expectedTag:   "latest",
+			expectedError: false,
+		},
+		{
+			ref:           "test.com:5000/test/repository",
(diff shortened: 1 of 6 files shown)


types/reference/image_reference_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"testing"
+)
+
+func TestParse(t *testing.T) {
+	testCases := []struct {
+		ref           string
+		expectedName  string
+		expectedTag   string
```
stevvooe: I'm not sure we should reinforce the concept of a tag being a digest. Maybe call this `expectedTagOrDigest`.

vdemeester: Ah yes :wink:

stevvooe:
LGTM

This may be more subtle in the engine, but I think this works.

vdemeester:
@stevvooe It would also be possible to not change anything in the engine, just handle the `all` attribute in `engine-api` and append or not the default tag — but I feel it makes it clearer by passing all to engine (although it changes the API and might be tricky to handle daemon wise for backward compatibilites..).

stevvooe:
@vdemeester I agree.

The engine-api needs to focus on user intent, rather than expose the subtleties of the docker api.

HackToday:
So it means change API, and allow to pass all=1 for `createimages` ? Is that ?

vdemeester:
@HackToday there is two solution, one with an API change, one without. I think we'll split it into two, one that fixes the behaviour without the API change, and then proposing the API change.

vdemeester:
Updated to require no Server API changes for now :angel:

vdemeester:
Updated by also modifying push behavior.

/cc @stevvooe @tonistiigi

calavera:
LGTM

LK4D4:
LGTM

PR #211, opened 2016-04-25
Remove omitempty from AuthConfig.Email

kencochrane:
In issue https://github.com/docker/docker/issues/22165 when the email field is missing from the auth config in `~/.docker/config.json` it is causing problems with older versions of docker-compose.

This fix removes `omitempty` from AuthConfig.Email, so that we write an empty Email field to the config.json file. I'm not 100% sure this will fix the issue with compose. 

This PR is mostly to start the discussion, if it looks good to people, we can move forward.

Signed-off-by: Ken Cochrane <KenCochrane@gmail.com>

### DIFF
--- a/types/auth.go
+++ b/types/auth.go
@@ -9,7 +9,7 @@ type AuthConfig struct {
 	// Email is an optional value associated with the username.
 	// This field is deprecated and will be removed in a later
 	// version of docker.
-	Email string `json:"email,omitempty"`
+	Email string `json:"email"`
 
 	ServerAddress string `json:"serveraddress,omitempty"`
 


thaJeztah:
ping @dnephin @calavera wdyt?

dnephin:
We fixed the logic in `docker-py` (which fixes Compose), so new versions of Compose will be fine. 

Even if we make this change now, Docker 1.11 will continue to not work with older versions of Compose. I'm not against making this change, but I don't know if it's totally necessary. It should be easy for users to upgrade Compose because it's just a client.

thaJeztah:
Alright, let's close this one, thanks for looking @dnephin

PR #227, opened 2016-05-18
Use filters.ToParamWithVersion elsewhere

vdemeester:
`ToParamWithVersion` allows to serialize correctly filters for oldeversion of docker (1.9.1 and lower mainly). It is currently only use for `ContainerList` but it's not the only place where the filters happens 🐰.

This make use of `ToParamWithVersion` for `ImageList`, `VolumesList`, `NetworkList` and `Events` as well (all where filters where already there in 1.9.1).

/cc @calavera @stevvooe @LK4D4 @runcom @thaJeztah

It's the reason behind https://github.com/containous/traefik/issues/360 and I'm pretty sure it could be reproduced using the official docker client. 

🐸

### DIFF
--- a/client/events.go
+++ b/client/events.go
@@ -33,7 +33,7 @@ func (cli *Client) Events(ctx context.Context, options types.EventsOptions) (io.
 		query.Set("until", ts)
 	}
 	if options.Filters.Len() > 0 {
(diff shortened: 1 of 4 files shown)


thaJeztah:
LGTM

calavera:
LGTM

Issue #207, opened 2016-04-22
Network need expose an interface NetworkInspectWithRaw


HackToday:
Right now, the docker issue https://github.com/docker/docker/issues/22231
because of engine-api 
```
NetworkInspect(ctx context.Context, networkID string) (types.NetworkResource, error)
```
it need a new interface like container/image inspect, defined as following
```
NetworkInspectWithRaw(ctx context.Context, networkID string) (types.NetworkResource, []byte, error)
```
Then the inspect can -f (go template) with `Id` like container/image did

HackToday:
#dibs

HackToday:
ping @calavera and @vdemeester  WDYT.

vdemeester:
@HackToday Sorry for the delay.. Wanted to answer before but got.. carried away.
Looking at the docker/docker issue, I think it makes sense to be consistent (having those methods for Containers, Images, Network and Volumes).

So I would definitely be ok with it. I also think, we might want to make sure all these 4 (container, image, network and volume) have both methods.

calavera:
yep, that's the way to go. You can read about the issue with Go templates further in this issue:

https://github.com/docker/go/pull/3

thaJeztah:
wow, that's one massive PR; sad it didn't get any attention

thaJeztah:
@HackToday did you want to work on this issue?

HackToday:
sure @thaJeztah  will check it , Miss that those days :0

HackToday:
@thaJeztah  One question for that, as this depend on engine-api, should I send a PR for engine-api, after that merged, then I send PR to docker-engine ? 
Do we need some vendor to do that ?

thaJeztah:
@HackToday I think in this case, it's good to open a PR here first, and open a PR to docker/docker after its merged here. Can you do PRs for all types? (Networks, Volumes etc) https://github.com/docker/engine-api/issues/207#issuecomment-215022955

PR #228, opened 2016-05-19
Add Raw interface for network,volume inspect

HackToday:
Closes: #207

### DIFF
--- a/client/volume_inspect.go
+++ b/client/volume_inspect.go
@@ -1,7 +1,9 @@
 package client
 
 import (
+	"bytes"
 	"encoding/json"
+	"io/ioutil"
 	"net/http"
 
 	"github.com/docker/engine-api/types"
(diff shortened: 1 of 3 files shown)


HackToday:
ping @vdemeester and @thaJeztah  PTAL let me know if it ok or not for this. Thanks

client/volume_inspect.go
```diff
@@ -22,3 +24,24 @@ func (cli *Client) VolumeInspect(ctx context.Context, volumeID string) (types.Vo
 	ensureReaderClosed(resp)
 	return volume, err
 }
+
+// VolumeInspectWithRaw returns the information about a specific volume in the docker host and it's raw representation
+func (cli *Client) VolumeInspectWithRaw(ctx context.Context, volumeID string) (types.Volume, []byte, error) {
```
dnephin: To remove the code duplicate with `VolumeInspect()` maybe `VolumeInspect()` should call this new method, and just discard the body?

This method does a little bit of extra work, but it shouldn't be significant, and the old method is really just there for backwards compatibility now (it won't actually get called by the engine I believe).

Same comment about the network inspect call.

vdemeester: @HackToday I definitely agree with @dnephin. I even think we could do this for the existing ones (in `container_inspect.go` and `image_inspect.go`).

HackToday:
@dnephin  and @vdemeester  PTAL the new change. Thanks

vdemeester:
LGTM 🐮

calavera:
LGTM

PR #233, opened 2016-05-21
Handle JSON errors from the server

bfirsh:
See https://github.com/docker/docker/pull/22880

### DIFF
--- a/client/request.go
+++ b/client/request.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/docker/engine-api/client/transport/cancellable"
+	"github.com/docker/engine-api/types"
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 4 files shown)


cpuguy83:
Hmm, I'm not sure about this.
I like how the registry is handling this with errcode, however I don't think I see much usefulness out of json errors with an HTTP status code and a message.

calavera:
This is also importing packages from `docker/docker`, which goes against the practices in this repository.

bfirsh:
@cpuguy83 There is a need for JSON errors of some kind, no strong feelings about whether there's a code or not. Some discussion of that in https://github.com/docker/docker/pull/22880

@calavera Ah, understood. I'll put the error struct in here.

bfirsh:
Added `ErrorResponse` instead of using `jsonmessage.APIError` from `docker/docker`.

client/request.go
```diff
@@ -130,7 +131,19 @@ func (cli *Client) sendClientRequest(ctx context.Context, method, path string, q
 		if len(body) == 0 {
 			return serverResp, fmt.Errorf("Error: request returned %s for API route and version %s, check if the server supports the requested API version", http.StatusText(serverResp.statusCode), req.URL)
 		}
-		return serverResp, fmt.Errorf("Error response from daemon: %s", bytes.TrimSpace(body))
+
+		var errorMessage string
+		if strings.Contains(resp.Header.Get("Content-Type"), "application/json") {
```
calavera: should this be `==`? what are the possible content types? we should check specifically for them.

bfirsh: The server returns `application/json; charset=utf-8`, and the default is `utf-8` so we should support just `application/json` too.

Realistically we'll only consume/produce utf-8, so perhaps this should check for the exact strings `application/json` or `application/json; charset=utf-8`.

calavera: SGTM

bfirsh: Hmm. Seems like I'm wrong and the Engine actually just returns `application/json`. Not sure where I found `charset=utf-8`. Perhaps it was in some documentation somewhere.

I've updated to just check for the exact string `application/json`.

calavera:
LGTM

justincormack:
Tested this and it seems to behave as I would expect with different API versions.

LGTM

(I would prefer to add the error code, but the whole idea is this is extensible later eg we can add stack traces in debug mode or line the error is from, or nested error info from runc etc later so lets go with this now).

thaJeztah:
moving this to code review; or were we there already?

justincormack:
Yes, code LGTM.

thaJeztah:
LGTM

Issue #230, opened 2016-05-19
ContainerStop and ContainerRestart should use time.Duration for timeout


aaronlehmann:
These API functions take an integer number of seconds for a timeout. It would be much more idiomatic Go to use `time.Duaration` to specify the timeout.

vdemeester:
ohhh I like that 😃

LK4D4:
I like that too

PR #235, opened 2016-05-25
Use time.Duration as type for timeouts

pdalpra:
This PR replaces instances of where timeout was an `int` representing the number of seconds by `time.Duration`, as suggested by #230 .
Not entirely sure about the float64 -> int conversion, also thought about using `strconv.FormatFloat(v, 'f', 0, 64)` for formatting it as an integer. WDYT ?

### DIFF
--- a/client/container_restart.go
+++ b/client/container_restart.go
@@ -2,17 +2,18 @@ package client
 
 import (
 	"net/url"
-	"strconv"
+	"time"
 
+	timetypes "github.com/docker/engine-api/types/time"
 	"golang.org/x/net/context"
 )
 
 // ContainerRestart stops and starts a container again.
 // It makes the daemon to wait for the container to be up again for
 // a specific amount of time, given the timeout.
-func (cli *Client) ContainerRestart(ctx context.Context, containerID string, timeout int) error {
+func (cli *Client) ContainerRestart(ctx context.Context, containerID string, timeout time.Duration) error {
 	query := url.Values{}
-	query.Set("t", strconv.Itoa(timeout))
+	query.Set("t", timetypes.DurationToSecondsString(timeout))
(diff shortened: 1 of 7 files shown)


vdemeester:
@pdalpra cool :+1: Using `strconv.FormatFloat` makes more sence to me (reading it at least).

pdalpra:
should I introduce this a utility function or inline the code for the time being ?

vdemeester:
@pdalpra maybe a utility function it `types/time` :stuck_out_tongue_closed_eyes:

pdalpra:
sorry for keeping it stale, will update the PR today

pdalpra:
@vdemeester PR (finally !) updated

vdemeester:
LGTM :angel: 
/cc @aaronlehmann @calavera

pdalpra:
I can take of the matching PR on `docker/docker` if you like

calavera:
LGTM

MHBauer:
I'm a bit late to the review. 

First, great thought and contribution @pdalpra. It is much appreciated and I think it is good practice to use `time.Duration`s for our timeout durations.


I do not like that we convert an int to a float then parse the float to convert to a string.

1. Too much conversion when it should be just number to number. (This is more of an aesthetics thing)
2. This adopts float semantics for an integer. (I think this is mainly to be used with full seconds, so it does not matter so much)

My suggestion would to be add a comment on the rounding, just so no-one expects to get sub second precision, or to adopt what is in the godocs for `time.Duration`, i.e. the calculation becomes `int64(duration/time.Second)` instead of the float/string parsing. That also does a rounding (always round down) with no sub second precision.

Thanks again @pdalpra for the contribution here and in docker/docker#23344!

PR #238, opened 2016-05-28
Make NotFound Errors exportable

vdemeester:
If they are not exportable, it's impossible to write unit-test on library using `engine-api` that would return an `imageNotFoundError` or other NotFoundErrors 🐹.

This is a naive approach and I'm not sure this is the right one either. But somehow, in some unit-test in `libcompose` I need to make my client returning a `networkNotFoundError` (or the same for other types) and there is currently no way I can do that (or I'm missing something :stuck_out_tongue_closed_eyes: ).

/cc @calavera @LK4D4 @runcom @stevvooe 

🐸

### DIFF
--- a/client/errors.go
+++ b/client/errors.go
@@ -8,26 +8,47 @@ import (
 // ErrConnectionFailed is an error raised when the connection between the client and the server failed.
 var ErrConnectionFailed = errors.New("Cannot connect to the Docker daemon. Is the docker daemon running on this host?")
 
+type notFound interface {
+	error
+	NotFound() bool // Is the error a NotFound error
+}
+
+// IsErrNotFound returns true if the error is caused with an
+// object (image, container, network, volume, …) is not found in the docker host.
+func IsErrNotFound(err error) bool {
+	te, ok := err.(notFound)
+	return ok && te.NotFound()
+}
+
 // imageNotFoundError implements an error returned when an image is not in the docker host.
 type imageNotFoundError struct {
 	imageID string
 }
 
+// NoFound indicates that this error type is of NotFound
+func (e imageNotFoundError) NotFound() bool {
+	return true
+}
+
 // Error returns a string representation of an imageNotFoundError
-func (i imageNotFoundError) Error() string {
-	return fmt.Sprintf("Error: No such image: %s", i.imageID)
+func (e imageNotFoundError) Error() string {
+	return fmt.Sprintf("Error: No such image: %s", e.imageID)
 }
 
 // IsErrImageNotFound returns true if the error is caused
 // when an image is not found in the docker host.
 func IsErrImageNotFound(err error) bool {
-	_, ok := err.(imageNotFoundError)
-	return ok
+	return IsErrNotFound(err)
 }
 
 // containerNotFoundError implements an error returned when a container is not in the docker host.
 type containerNotFoundError struct {
 	containerID string
+}
+
+// NoFound indicates that this error type is of NotFound
+func (e containerNotFoundError) NotFound() bool {
+	return true
 }
 
 // Error returns a string representation of a containerNotFoundError
@@ -38,13 +59,17 @@ func (e containerNotFoundError) Error() string {
 // IsErrContainerNotFound returns true if the error is caused
 // when a container is not found in the docker host.
 func IsErrContainerNotFound(err error) bool {
-	_, ok := err.(containerNotFoundError)
-	return ok
+	return IsErrNotFound(err)
 }
 
 // networkNotFoundError implements an error returned when a network is not in the docker host.
 type networkNotFoundError struct {
 	networkID string
+}
+
+// NoFound indicates that this error type is of NotFound
+func (e networkNotFoundError) NotFound() bool {
+	return true
 }
 
 // Error returns a string representation of a networkNotFoundError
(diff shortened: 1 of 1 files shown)


runcom:
There's an interesting read about errors: Assert errors for behaviour, not type http://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully

:smile_cat:

vdemeester:
@runcom yeah I've read it this last week, this morning too but… not sure how to apply there though…

Maybe like this : 

```go
type InspectNotFoundError interface {
    ID() string
}

type imageNotFoundError struct {
    imageID string
}

func (i imageNotFoundError) ID() string {
    return i.imageID
}

func (i imageNotFoundError) Error() string {
    return fmt.Sprintf("Error: No such image: %s", i.ImageID)
}

// …
func IsErrNotFound(err error) bool {
    _, ok := err.(InspectNotFoundError)
    return ok
}
```

… and thus removing all `IsErrContainerNotFound`, … and just keep a `IsErroNotFound`.

WDYT ?

thaJeztah:
Perhaps I'm misunderstanding, but weird thing is that Golang itself exports it's errors, e.g. [ErrNotExist](https://golang.org/pkg/os/#ErrNotExist), but looks like they do some additional trickery https://golang.org/src/os/error_unix.go?h=isNotExist#L25 (called from [`os.IsNotExist()`](https://golang.org/pkg/os/#IsNotExist)), so should we have a `IsNotFound()` function, that returns true if it's an ImageNotFound, VolumeNotFound etc? Sorry if I'm talking BS here :sweat_smile: 

Note; `IsErrInspectNotFound(err)` should probably be `IsInspectNotFound(err)` (no `Err` in the name)

calavera:
It makes sense that the errors are exported if we're going to generate them from other libraries. They were originally intended to be private.

LGTM

runcom:
I don't think we should generate those from other libraries either. Since those errors are "not found" errors, at least we should provide a way to check that an error is of type not found, not by exporting them (same as go/net does for OpError)

runcom:
And other libraries should not couple like this imo.

vdemeester:
@runcom I'll rework it with the interface, as it makes less/almost no coupling :wink:

runcom:
looks better to me - `NotFound()` seems not necessary though (except for the interface implementation) because it always returns true :/

client/errors.go
```diff
@@ -38,15 +52,20 @@ func (e containerNotFoundError) Error() string {
 // IsErrContainerNotFound returns true if the error is caused
 // when a container is not found in the docker host.
 func IsErrContainerNotFound(err error) bool {
-	_, ok := err.(containerNotFoundError)
-	return ok
+	te, ok := err.(notFound)
+	return ok && te.NotFound()
```
stevvooe: Can we centralize this?

vdemeester: @stevvooe hum what do you mean ? :sweat_smile:

runcom: probably a generic notfound handler which does type assertion + ok && ...?

stevvooe:
LGTM

May want to introduce a separate `errors` package to centralize error handling but that is outside the scope of this PR.

Prevents this in client code:

```
func isUnknownContainer(err error) bool {
	return strings.Contains(err.Error(), "No such container:")
}
```

@runcom 
> looks better to me - NotFound() seems not necessary though

Indeed, but it represents the [best practices in error handling](http://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully). I don't agree with everything in that blog post, but it's extremely composable and highly flexible. See `net.Temporary/net.Timeout` for another example.

runcom:
> Indeed, but it represents the best practices in error handling. I don't agree with everything in that blog post, but it's extremely composable and highly flexible. See net.Temporary/net.Timeout for another example.

me neither - was just pointing that out thinking out loud - `net.Temporary/net.Timeout` is exactly what I was looking before chiming in here ;)

vdemeester:
Updated
/cc @runcom @LK4D4 @calavera @stevvooe :angel:

client/errors.go
```diff
@@ -72,8 +101,7 @@ func (e volumeNotFoundError) Error() string {
 // IsErrVolumeNotFound returns true if the error is caused
 // when a volume is not found in the docker host.
 func IsErrVolumeNotFound(err error) bool {
-	_, ok := err.(volumeNotFoundError)
-	return ok
+	return IsErrNotFound(err)
```
runcom: I think we're missing something - is `IsErrNotFound` is exported any of the volume|image|container error not found will pass this check but there's no way to tell if it was volume|image|container - I'm suggesting to have a filed (like `Op` filed in go net error if I'm not wrong) which tell us more about the error. Does this make sense? /cc @stevvooe

vdemeester: Hum good point :+1:

stevvooe: The nice thing about this approach is that we don't necessarily need to add it in this PR. We just define a new interface method that exports the object type.

Is there a direct need for this as part of this PR?

stevvooe:
LGTM

client/errors.go
runcom: I don't think so (even if it would be good, I guess a follow up is fine though)

runcom:
LGTM

calavera:
Still LGTM

PR #240, opened 2016-05-30
Only set url value when it's not empty

WeiZhang555:
Only a small nit.
Currently docker start API always contains a trailing `checkpoint=` even
if value is empty, this is unnecessary, we can set it only when it's not
empty.

### DIFF
--- a/client/container_start.go
+++ b/client/container_start.go
@@ -4,12 +4,16 @@ import (
 	"net/url"
 
 	"golang.org/x/net/context"
+
+	"github.com/docker/engine-api/types"
 )
 
 // ContainerStart sends a request to the docker daemon to start a container.
-func (cli *Client) ContainerStart(ctx context.Context, containerID string, checkpointID string) error {
+func (cli *Client) ContainerStart(ctx context.Context, containerID string, options types.ContainerStartOptions) error {
 	query := url.Values{}
-	query.Set("checkpoint", checkpointID)
+	if len(options.CheckpointID) != 0 {
+		query.Set("checkpoint", options.CheckpointID)
+	}
 
 	resp, err := cli.post(ctx, "/containers/"+containerID+"/start", query, nil, nil)
 	ensureReaderClosed(resp)
--- a/client/container_start_test.go
+++ b/client/container_start_test.go
@@ -10,13 +10,15 @@ import (
 	"testing"
 
 	"golang.org/x/net/context"
+
+	"github.com/docker/engine-api/types"
 )
 
 func TestContainerStartError(t *testing.T) {
 	client := &Client{
 		transport: newMockClient(nil, errorMock(http.StatusInternalServerError, "Server error")),
 	}
-	err := client.ContainerStart(context.Background(), "nothing", "")
+	err := client.ContainerStart(context.Background(), "nothing", types.ContainerStartOptions{})
 	if err == nil || err.Error() != "Error response from daemon: Server error" {
 		t.Fatalf("expected a Server Error, got %v", err)
 	}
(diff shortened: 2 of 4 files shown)


client/container_start.go
```diff
@@ -9,7 +9,9 @@ import (
 // ContainerStart sends a request to the docker daemon to start a container.
 func (cli *Client) ContainerStart(ctx context.Context, containerID string, checkpointID string) error {
 	query := url.Values{}
-	query.Set("checkpoint", checkpointID)
+	if len(checkpointID) != 0 {
```
vdemeester: Hum if `checkpointID` is optional (which seems it is), then I would definitely move it to a `Option` struct, even for one fiead… @stevvooe @calavera WDYT ?

vdemeester:
One question / nit, other than that LGTM :+1:

thaJeztah:
ideally, it wouldn't send any parameter that is "default" (think there's still some other parameters that are sent even if no value is set by the user), so SGTM

client/container_start.go
calavera: SGTM

WeiZhang555: OK. So I'll move it to an `Option` struct, I can see that `checkpoint` is not in use in master branch, so I guess this won't break API.

vdemeester: Well, it gonna need some vendoring (right now we pass "").

WeiZhang555: I'll do the vendoring later after this. I'm planning to check other APIs, to see if they have same problem. BUT, I really hope not to introduce unnecessary break to existing APIs.  :smile:

WeiZhang555:
Updated,  I didn't touch some very old APIs in case to break anything.
PTAL.  :smile:

client/checkpoint_delete.go
```diff
@@ -2,11 +2,13 @@ package client
...
 	"golang.org/x/net/context"
+
+	"github.com/docker/engine-api/types"
 )
 
 // CheckpointDelete deletes the checkpoint with the given name from the given container
-func (cli *Client) CheckpointDelete(ctx context.Context, containerID string, checkpointID string) error {
-	resp, err := cli.delete(ctx, "/containers/"+containerID+"/checkpoints/"+checkpointID, nil, nil)
+func (cli *Client) CheckpointDelete(ctx context.Context, containerID string, options types.CheckpointDeleteOptions) error {
+	resp, err := cli.delete(ctx, "/containers/"+containerID+"/checkpoints/"+options.CheckpointID, nil, nil)
```
stevvooe: This treats `/containers/<containerID>/checkpoints/` and `/containers/<containerID>/checkpoints`, when it is not. This should probably use a url-safe join function.

vdemeester: @WeiZhang555 I think `checkpointID` on `CheckpointDelete` is required, so it should not be changed in an options (just `ContainerStart` imho).

WeiZhang555: Ah, right. This shouldn't be considered as an option. Let me change it back.

WeiZhang555: @stevvooe Agree, I guess at least we should validate the checkpointID string.
There're some more in other APIs.

WeiZhang555:
Updated

runcom:
LGTM do we need to change something in docker/docker when re-vendoring there?

vdemeester:
@runcom yes a small thing where `ContainerStart` is called :angel:. Given the fact that `checkpointID` is optional we should have make the *option* struct first time but meh :stuck_out_tongue_closed_eyes:.

Still LGTM :+1:

WeiZhang555:
Re-vendoring is enough, we haven't put this API

calavera:
LGTM

WeiZhang555:
I'll do a vendoring to make it work in `docker/docker` :smile:

vdemeester:
@WeiZhang555 I oprned a PR on docker/docker for it yesterday (Bump engine-api ...) :stuck_out_tongue_closed_eyes:

PR #247, opened 2016-06-01
Add ipam-info in 'network inspect' response

vikaschoudhary16:
Related to https://github.com/docker/docker/pull/23163

Signed-off-by: vikaschoudhary16 <choudharyvikas16@gmail.com>

### DIFF
--- a/types/network/network.go
+++ b/types/network/network.go
@@ -11,6 +11,7 @@ type IPAM struct {
 	Driver  string
 	Options map[string]string //Per network IPAM driver options
 	Config  []IPAMConfig
+	Info    []IPAMInfo
 }
 
 // IPAMConfig represents IPAM configurations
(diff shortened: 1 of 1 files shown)


vikaschoudhary16:
Cc @mavenugo @aboch @mrjana @MoBani

types/network/network.go
```diff
@@ -21,6 +22,13 @@ type IPAMConfig struct {
 	AuxAddress map[string]string `json:"AuxiliaryAddresses,omitempty"`
 }
 
+// IPAMInfo represents IPAM operational information
+type IPAMInfo struct {
+	PoolID  string `json:",omitempty"`
```
aboch: `PoolID` is an index the ipam driver provides to libnetwork in response to pool requests for future lookups when requesting/releasing addresses.
It is not an information for the docker end user/app. It is not meant for that, it is not part of the operational information. I don't think it should be included in this structure.

vikaschoudhary16: @aboch ,
Many thanks for comment.

As you suggested here https://github.com/docker/docker/issues/23025, for handling identical cidr pools, address-spaces can be used but many networking backends still have problems in supporting address-spaces. In that case, these drivers can invoke "network inspect" rest api to get PoolID and pick the correct pool.

Also showing PoolID will provide more visiblity to user, in case user is aware of IPAM pools.

In case your concern is just not to use "Info" word for PoolID as it is not an operational info, Will it work for you if struct is renamed to something this "IpamData"?

Looking forward for your valuable inputs.

cpuguy83:
Closing since the mentioned docker PR was closed

PR #250, opened 2016-06-01
Add `--limit` option to `docker search`

yongtang:
This fix is related to docker:
https://github.com/docker/docker/pull/23107
https://github.com/docker/docker/issues/23055

Currently docker search result caps at 25 and there is no way to allow getting more results (if exist).

This fix adds the flag --limit so that it is possible to return more results from the docker search.

### DIFF
--- a/client/image_search.go
+++ b/client/image_search.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"encoding/json"
+	"fmt"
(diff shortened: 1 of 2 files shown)


vdemeester:
LGTM 🐼

thaJeztah:
LGTM

PR #251, opened 2016-06-01
Add support for using different OCI runtimes

mlaventure:
Required by docker/docker#22983

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -252,6 +252,8 @@ type Info struct {
 	ClusterStore       string
 	ClusterAdvertise   string
 	SecurityOptions    []string
+	Runtimes           map[string]Runtime
+	DefaultRuntime     string
 }
 
 // PluginsInfo is a temp struct holding Plugins name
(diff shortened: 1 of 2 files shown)


vdemeester:
LGTM 🐶
/cc @calavera @MHBauer @vieux @LK4D4 @icecrime

types/types.go
```diff
@@ -252,6 +252,8 @@ type Info struct {
 	ClusterStore       string
 	ClusterAdvertise   string
 	SecurityOptions    []string
+	Runtimes           map[string]Runtime
```
icecrime: Should we expose the path in the `/info` output?

mlaventure: We do:

```
# docker info
...
Runtimes:
 oci: /usr/local/bin/docker-runc (--debug)
 default: docker-runc
 foo: /usr/local/bin/foo (--debug)
Default Runtime: foo
...
```

It's part of the Runtime struct, we also display the runtime arguments if any were defined

calavera:
LGTM

types/types.go
icecrime: Sorry my comment was ambiguous: I meant I'm not sure we should expose the path. But I guess it shouldn't hurt, let's do it :+1:

PR #255, opened 2016-06-03
Remove /json from API URLs

bfirsh:
See https://github.com/docker/docker/pull/23219

### DIFF
--- a/client/container_exec.go
+++ b/client/container_exec.go
@@ -38,7 +38,7 @@ func (cli *Client) ContainerExecAttach(ctx context.Context, execID string, confi
 // ContainerExecInspect returns information about a specific exec process on the docker host.
 func (cli *Client) ContainerExecInspect(ctx context.Context, execID string) (types.ContainerExecInspect, error) {
 	var response types.ContainerExecInspect
-	resp, err := cli.get(ctx, "/exec/"+execID+"/json", nil, nil)
+	resp, err := cli.get(ctx, "/exec/"+execID, nil, nil)
 	if err != nil {
 		return response, err
 	}
--- a/client/container_exec_test.go
+++ b/client/container_exec_test.go
@@ -124,7 +124,7 @@ func TestContainerExecInspectError(t *testing.T) {
 }
 
 func TestContainerExecInspect(t *testing.T) {
-	expectedURL := "/exec/exec_id/json"
+	expectedURL := "/exec/exec_id"
 	client := &Client{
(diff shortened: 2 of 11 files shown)


calavera:
LGTM

bfirsh:
FWIW this has a hard dependency on https://github.com/docker/docker/pull/23219, and that has stalled on a design decision.

vdemeester:
Moving to code-review as docker/docker#23219 is in code-review

vdemeester:
LGTM 🐹

calavera:
still LGTM

tamird:
So...this makes this library not work with older docker versions? For example, the latest version offered by CircleCI is 1.10.

vdemeester:
@tamird arf, yeah that is pretty bad :sweat: 
I'll revert this for now and we'll find a better way to handle it

tamird:
Tim BL had it right all the way back in 1998 https://www.w3.org/Provider/Style/URI.html

vdemeester:
@tamird the old URL won't change (probably for the next… 5 years or eternity) but the client code might… although we didn't do right… Sorry about that :sweat:

tamird:
This client code should also not change until all docker versions that don't support these URLs are EOL'd.

Making this change now is breaking compatibility for no benefit whatsoever, especially given the state of dependency management in Go.

vdemeester:
@tamird see https://github.com/docker/engine-api/pull/263 :bow:

tamird:
👍  thanks

bfirsh:
My bad. I should have closed this while there was still a design decision in https://github.com/docker/docker/pull/23219.

I wonder if we should open the discussion about how to version engine-api. People should presumably be locking to API versions somehow rather than using master. (See also #260.)

bfirsh:
Ah, nevermind, I see there is `cli.version`, so I should have used that in this PR to check what URL to make the request to.

Perhaps we should set the default version to the current stable version of Docker so people using engine-api off master don't get broken by default.

vdemeester:
> Perhaps we should set the default version to the current stable version of Docker so people using engine-api off master don't get broken by default.

That's a good idea :angel:

bfirsh:
@vdemeester I suppose you have to explicitly pass a blank string to `client.NewClient` if you don't want to set a version, so I guess that's pretty much saying "I expect this to break". I'll add a warning to the godoc that you really should be specifying a version.

It's different for `NewEnvClient` though: #264

PR #256, opened 2016-06-03
Add config parameter to change per-container stop timeout during daemon shutdown

yongtang:
This fix tries to add a `StopTimeout` in container's Config so that a per-container stop timeout could be specified during daemon shutdown.

The `StopTimeout` is defined as `*int` so that unspecified value could be correctly identified as using the default value (10s).
Note: Both `0` and `-1` could not be used as the default value because `0` means stop immediately and `-1` means wait forever.

This fix is related to docker:
https://github.com/docker/docker/pull/22566


vdemeester:
LGTM 🐮

coolljt0725:
LGTM

stevvooe:
"Make the zero-value useful"

Could we please stop adding needless pointer values everywhere? This is getting tiresome. `1ns` is just as good as zero and doesn't cause all kinds of havoc.

PR #258, opened 2016-06-03
Add new APIs for safer mount specifications

cpuguy83:
This is for docker/docker#22373

### DIFF
--- a/types/container/host_config.go
+++ b/types/container/host_config.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	"github.com/docker/engine-api/types/blkiodev"
+	"github.com/docker/engine-api/types/mount"
 	"github.com/docker/engine-api/types/strslice"
 	"github.com/docker/go-connections/nat"
 	"github.com/docker/go-units"
@@ -317,4 +318,7 @@ type HostConfig struct {
 
 	// Contains container's resources (cgroups, ulimits)
 	Resources
+
+	// Mounts specs used by the container
+	Mounts []mount.Mount `json:",omitempty"`
 }
--- a/types/types.go
+++ b/types/types.go
@@ -5,6 +5,7 @@ import (
 	"time"
 
 	"github.com/docker/engine-api/types/container"
+	"github.com/docker/engine-api/types/mount"
 	"github.com/docker/engine-api/types/network"
 	"github.com/docker/engine-api/types/registry"
 	"github.com/docker/engine-api/types/swarm"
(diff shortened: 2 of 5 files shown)


cpuguy83:
ping @thaJeztah @justincormack

cpuguy83:
omg `go lint`....

MHBauer:
Remove the volume structs from the base `types` package as well, or leave it there for compatibility reasons? Clients will have to change to the new package.

cpuguy83:
IIRC I moved these around due to circular imports.

cpuguy83:
Yes, so issue is `types/container` pkg needs volume create struct, which caused an import to `types` pkg, and `types` is already importing `types/container`... so the volume `CreateRequest` needs to be in a separate package.
Automatically I'm going to bring over `ListResponse` to that new package b/c these should not be split up, which means I also need to bring over the `Volume` struct or else have another circular import... `types/container` -> `types/volume` -> `types`

cpuguy83:
Lint issues and all are fixed, CI is passing now.

types/container/host_config.go
```diff
@@ -317,4 +334,7 @@ type HostConfig struct {
 
 	// Contains container's resources (cgroups, ulimits)
 	Resources
+
+	// Mounts sepcs used by the container
```
thaJeztah: s/sepcs/specs/

thaJeztah:
LGTM

calavera:
@cpuguy83 this is modifying the existing endpoints, does that mean that the docker cli is going to use them too?

cpuguy83:
@calavera Not sure I follow.

calavera:
oh, nevermind, I should have read more closely.

cpuguy83:
This is now updated and uses (or rather moves) the swarm mount types.

thaJeztah:
re-LGTM

types/types.go
```diff
@@ -410,13 +411,14 @@ type DefaultNetworkSettings struct {
 
 // MountPoint represents a mount point configuration inside the container.
 type MountPoint struct {
-	Name        string `json:",omitempty"`
+	Type        mount.Type `json:",omitempty"`
+	Name        string     `json:",omitempty"`
 	Source      string
 	Destination string
```
stevvooe: Wish we could alias these fields. I'm guessing this is exported via docker info or docker inspect?

stevvooe:
LGTM

I like that we've combined these.

types/types.go
cpuguy83: Yes.

cpuguy83:
With 2 LGTMs I'm merging.
Thanks!

PR #264, opened 2016-06-07
Add default version to NewEnvClient

bfirsh:
Matches behaviour of CLI, and will stop apps breaking if they're running off engine-api master, but not docker/docker master.

See also #255.


thaJeztah:
LGTM

bfirsh:
Updated the default version to 1.23, because I guess that's the current stable version. :)

vdemeester:
LGTM 😇

PR #265, opened 2016-06-08
Add swarm management API endpoints

vieux:
As described in our [ROADMAP.md](https://github.com/docker/docker/blob/master/ROADMAP.md#14-cluster-capable-engine), introduce new swarmkit management API endpoints and types.

The new concepts introduced are:
* Swarm: a group of Docker Engines collaborating to schedule services.
* Nodes: Engines participating in a Swarm.
* Services: Swarm managed entities.
* Tasks: unit of scheduling (a service gives birth to multiple task, which in our case corresponds to running containers).


client/interface.go
```diff
@@ -18,6 +19,22 @@ type APIClient interface {
 	CheckpointCreate(ctx context.Context, container string, options types.CheckpointCreateOptions) error
 	CheckpointDelete(ctx context.Context, container string, checkpointID string) error
 	CheckpointList(ctx context.Context, container string) ([]types.Checkpoint, error)
+	SwarmInit(ctx context.Context, req swarm.InitRequest) (string, error)
```
vdemeester: The `APIClient` interface is getting bigger and bigger. We might want to split it into several ones (like `VolumeAPIClient`, `NetworkAPIClient`, `SwarmAPIClient`, …) and have `APIClient` embedded them… 

It's a nit and not in the scope of the PR… just thinking outloud.

*This would help project using `engine-api` to have less work to do to unit-test some part :angel:*

@calavera @vieux @icecrime @stevvooe WDYT ?

types/swarm/container.go
```diff
@@ -0,0 +1,74 @@
...
+	Args            []string          `json:",omitempty"`
+	Env             []string          `json:",omitempty"`
+	Dir             string            `json:",omitempty"`
+	User            string            `json:",omitempty"`
+	Mounts          []Mount           `json:",omitempty"`
+	StopGracePeriod *time.Duration    `json:",omitempty"`
+}
+
+// Mount represents a mount (volume).
+type Mount struct {
```
vdemeester: There is a little bit of redundancy with `types.Volume`, `types.MountPoint` in current master and what about #258 ? /cc @cpuguy83

types/swarm/network.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// IPAMOptions represents ipam options.
+type IPAMOptions struct {
+	Driver  Driver       `json:",omitempty"`
+	Configs []IPAMConfig `json:",omitempty"`
+}
+
+// IPAMConfig represents ipam configuration.
+type IPAMConfig struct {
```
vdemeester: Same here and `types/network` package. I think everything that is common between `swarm` and *standalone* engine could be elsewhere

client/swarm.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"encoding/json"
+	"net/url"
+
+	"github.com/docker/engine-api/types/swarm"
+	"golang.org/x/net/context"
+)
+
+// SwarmInit initializes the Swarm.
+func (cli *Client) SwarmInit(ctx context.Context, req swarm.InitRequest) (string, error) {
```
vdemeester: This is probably the *ultimate* nit but, all other methods have their own file, why not swarm ones ? :angel: :stuck_out_tongue_closed_eyes:

types/swarm/container.go
cpuguy83: I wonder how that happened :)

It's probably fine to merge these today, but a swarm `Mount` may indeed end up with different requirements in the future.

I also took out `MCSAaccessMode` from mine, we might want to do it here as well.

IIRC, swarm does not let you schedule a task to a node that doesn't have a bind path, and as such it doesn't make sense to allow tweaking the selinux label.

client/interface.go
dnephin: Yes, we should definitely do this.

vdemeester: Cool, I'll prepare a follow-up for this then :wink:

vieux:
thanks for the review @vdemeester I'll push some changes soon

crosbymichael:
LGTM

nice looking code +1

types/swarm/container.go
stevvooe: @vdemeester These are still being reconciled. We really need to remove the confusion between mounts and volumes.

types/swarm/network.go
stevvooe: This is the current `IPAMConfig`:

```
// IPAMConfig represents IPAM configurations
type IPAMConfig struct {
	Subnet     string            `json:",omitempty"`
	IPRange    string            `json:",omitempty"`
	Gateway    string            `json:",omitempty"`
	AuxAddress map[string]string `json:"AuxiliaryAddresses,omitempty"`
}
```

Does anybody know what an `AuxAddress` is? Anyone?
No. These are reserved addresses.

IPRange? O, I was confused, I thought we were talking about oxen.

types/swarm/network.go
```diff
@@ -0,0 +1,121 @@
...
+	// IPAMConfigFamilyUnknown UNKNOWN
+	IPAMConfigFamilyUnknown = "UNKNOWN"
+	// IPAMConfigFamilyIpv4 IPV4
+	IPAMConfigFamilyIpv4 = "IPV4"
+	// IPAMConfigFamilyIpv6 IPV6
+	IPAMConfigFamilyIpv6 = "IPV6"
+)
+
+// IPAMConfigFamily represents the family of a ipam configuration.
+type IPAMConfigFamily string
```
stevvooe: Family can be detected by the format of subnet. Can we remove this?

@mrjana

types/swarm/network.go
stevvooe: `Reserved map[string]string` -> I still have no clue what is on the other side of this map.

stevvooe: Let's remove `Reserved`, as well. There is absolutely no documentation on how this works or what it is supposed to be.

types/swarm/network.go
mrjana: Yes we can remove that.

vieux: @mrjana ok I'm removing it, can you remove it from swarmkit ?

types/swarm/network.go
mrjana: This is only used in the `Network` object which is embedded in `NetworkAttachment`. So we can remove `Reserved` here. No problems with backward compatibility.

icecrime:
LGTM

vdemeester:
ah golint erro

```
types/swarm/container.go:57:6: exported type BindOptions should have comment or be unexported
```

types/swarm/network.go
```diff
@@ -0,0 +1,99 @@
...
+	DriverConfiguration *Driver      `json:",omitempty"`
+	IPv6Enabled         bool         `json:",omitempty"`
+	Internal            bool         `json:",omitempty"`
+	IPAMOptions         *IPAMOptions `json:",omitempty"`
+}
+
+// NetworkAttachmentConfig represents the configuration of a network attachement.
+type NetworkAttachmentConfig struct {
+	Target  string   `json:",omitempty"`
+	Aliases []string `json:",omitempty"`
```
vieux: ping @mavenugo  @icecrime

types/swarm/swarm.go
```diff
@@ -0,0 +1,107 @@
...
+	Spec            Spec
+}
+
+// JoinRequest is the request used to join a swarm.
+type JoinRequest struct {
+	ListenAddr string
+	RemoteAddr string
+	Secret     string // accept by secret
+	CACertHash string
+	Manager    bool
```
stevvooe: This isn't quite right. What is this used for?

types/swarm/swarm.go
```diff
@@ -0,0 +1,107 @@
...
+type InitRequest struct {
+	ListenAddr      string
+	ForceNewCluster bool
+	Spec            Spec
+}
+
+// JoinRequest is the request used to join a swarm.
+type JoinRequest struct {
+	ListenAddr string
+	RemoteAddr string
```
stevvooe: This should be a list of manager addresses for the node to join.

types/swarm/network.go
icecrime: :+1:

types/swarm/swarm.go
vieux: `docker swarm join <ip:port> --manager`

/cc @tonistiigi

types/swarm/swarm.go
vieux: changed to `RemoteAddrs []string`

client/interface.go
```diff
@@ -19,6 +20,22 @@ type APIClient interface {
...
+	SwarmJoin(ctx context.Context, req swarm.JoinRequest) error
+	SwarmLeave(ctx context.Context, force bool) error
+	SwarmInspect(ctx context.Context) (swarm.Swarm, error)
+	SwarmUpdate(ctx context.Context, swarm swarm.Swarm) error
+	NodeInspect(ctx context.Context, nodeID string) (swarm.Node, error)
+	NodeList(ctx context.Context, options types.NodeListOptions) ([]swarm.Node, error)
+	NodeRemove(ctx context.Context, nodeID string) error
+	NodeUpdate(ctx context.Context, nodeID string, node swarm.Node) error
```
stevvooe: This needs to take `swarm.NodeSpec`.

icecrime: Actually that's probably true (cc @vieux @dnephin)

client/interface.go
```diff
@@ -19,6 +20,22 @@ type APIClient interface {
 	CheckpointCreate(ctx context.Context, container string, options types.CheckpointCreateOptions) error
 	CheckpointDelete(ctx context.Context, container string, checkpointID string) error
 	CheckpointList(ctx context.Context, container string) ([]types.Checkpoint, error)
+	SwarmInit(ctx context.Context, req swarm.InitRequest) (string, error)
+	SwarmJoin(ctx context.Context, req swarm.JoinRequest) error
+	SwarmLeave(ctx context.Context, force bool) error
+	SwarmInspect(ctx context.Context) (swarm.Swarm, error)
+	SwarmUpdate(ctx context.Context, swarm swarm.Swarm) error
```
stevvooe: Should take `swarm.Spec`.

client/interface.go
vieux: I added a new commit, PTAL

icecrime:
LGTM :+1:

stevvooe:
LGTM 99🎈

icecrime:
LGTM :100:

tonistiigi:
LGTM

PR #271, opened 2016-06-14
Add config parameter to change per-container stop timeout during daemon shutdown

yongtang:
This fix tries to add config paramter to change per-contaienr stop timeout during the daemon shutdown. As the part of the update, this fix also changes `docker restart` and `docker stop` so that if `--time` is not specified, then the default stop timeout (10s) will be used.

This fix is related to
https://github.com/docker/docker/pull/22566
https://github.com/docker/docker/issues/22471

### DIFF
--- a/client/container_restart.go
+++ b/client/container_restart.go
@@ -11,9 +11,11 @@ import (
 // ContainerRestart stops and starts a container again.
 // It makes the daemon to wait for the container to be up again for
 // a specific amount of time, given the timeout.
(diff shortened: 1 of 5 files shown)


thaJeztah:
LGTM

calavera:
LGTM

stevvooe:
"Make the zero-value useful"

Could we please stop adding needless pointer values everywhere? This is getting tiresome. 1ns is just as good as zero and doesn't cause all kinds of havoc.

PR #277, opened 2016-06-14
Add live restore to Info struct

aboch:
This allows some validation during swarm init (see https://github.com/docker/docker/pull/23524 for details) and would allow the --live-restore option to be shown in the `docker info` o/p.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -256,6 +256,7 @@ type Info struct {
 	Runtimes           map[string]Runtime
 	DefaultRuntime     string
 	Swarm              swarm.Info
+	LiveRestore        bool
 }
 
 // PluginsInfo is a temp struct holding Plugins name


mavenugo:
LGTM

icecrime:
Thanks @aboch. I don't think the field needs to be in a public API for the daemon to take a decision accordingly. Can you please let me know if we don't have any other choice?

So far, not LGTM from my point of view.

aboch:
@icecrime The validation of swarm comamnd is done in the `cluster

aboch:
@icecrime Current compatibility validation for the swarm command is being done in the `cluster` pkg as a cluster method. Wanted to keep the validation in the same place. I am not sure of another smple option. Will check.

icecrime:
I'm thinking that:

- `executor.Backend` could have an additional `fun IsSwarmCompatible() bool` function
- The daemon can look into its internals (`--cluster-store`, `--live-restore`, or anything else really) to reply accordingly.

aboch:
Thanks for the suggestion @icecrime . Let me see if I can get it working.

aboch:
Thanks @icecrime I made the change and verified it works.
See https://github.com/docker/docker/pull/23524/files#diff-9cb7240401da846ac7c2252759538e9aR168

I think this PR can be closed, and new method addition review can happen in the docker/docker PR.

icecrime:
Nice! Thanks

PR #279, opened 2016-06-14
add omitempty to storageopts

vieux:
https://github.com/docker/engine-api/pull/151 broke backward compat, since `StorageOpt` is unused most of a time, let's put an `omitempty` to prevent unmarshalling error.

ping @alexmavr


alexmavr:
LGTM, however if the struct is used wouldn't the unmarshalling error still occur?

vieux:
@alexmavr true but fixing this takes more time, I think this one would help us in the meantime.

alexmavr:
+1, thanks again for your attention on this

stevvooe:
LGTM

This is the right fix.

vdemeester:
LGTM 🐮

PR #280, opened 2016-06-15
Update PluginInstall signature for consistency

vdemeester:
This make PluginInstall a little bit more consistent with the rest of the api code base 🐮:

- Make `enable` and `acceptAllPermissions` optional
- Adds `PrivilegeFunc` to make it ask for login if the daemon error is an   HTTP Unauthorized (just like image pull/push).
- Add a `AcceptPermissionsFunc` and remove Reader/Writer from the   signature. The idea is to not force interactive input/output and let   the user of the API defines the behavior for accepting permissions.

/cc @thaJeztah @runcom @tiborvass @calavera @stevvooe @icecrime 

Related PR : https://github.com/docker/docker/issues/23570

🐸

### DIFF
--- a/types/plugin.go
+++ b/types/plugin.go
@@ -7,6 +7,15 @@ import (
 	"fmt"
 )
 
+// PluginInstallOptions holds parameters to install a plugin.
+type PluginInstallOptions struct {
+	Disabled              bool
+	AcceptAllPermissions  bool
+	RegistryAuth          string // RegistryAuth is the base64 encoded credentials for the registry
+	PrivilegeFunc         RequestPrivilegeFunc
+	AcceptPermissionsFunc func(PluginPrivileges) (bool, error)
+}
+
 // PluginConfig represents the values of settings potentially modifiable by a user
 type PluginConfig struct {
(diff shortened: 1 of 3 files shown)


calavera:
:clap: LGTM

types/plugin.go
```diff
@@ -7,6 +7,15 @@ import (
 	"fmt"
 )
 
+// PluginInstallOptions holds parameters to install a plugin.
+type PluginInstallOptions struct {
+	Enable                bool
```
stevvooe: This was "NoEnable" because it allowed one to control whether or not to enable the plugin immediately. If we want the default to be "enabled", this field should be `Disabled`.

stevvooe:
LGTM

A few comments around intent of `NoEnable`.

types/plugin.go
vdemeester: Ah true, make sense, I can update 😇

vdemeester:
@stevvooe Updated with `Disabled`

stevvooe:
LGTM

tiborvass:
LGTM

PR #281, opened 2016-06-15
Passing headers for service create/update

nishanttotla:
Currently, it is not possible to create or update services that require pulling of private images. This is a temporary fix to enable passing an auth token to SwarmKit from Docker. The token is retrieved at the client, and passed through as an `x-registry-auth` header which is then retrieved and added to `ContainerSpec` just before making a call to SwarmKit.

`engine-api` is in the call path, and needs to pass headers appropriately. Also see https://github.com/docker/docker/pull/23584

### DIFF
--- a/client/interface.go
+++ b/client/interface.go
@@ -100,11 +100,11 @@ type NodeAPIClient interface {
 
 // ServiceAPIClient defines API client methods for the services
 type ServiceAPIClient interface {
-	ServiceCreate(ctx context.Context, service swarm.ServiceSpec) (types.ServiceCreateResponse, error)
+	ServiceCreate(ctx context.Context, service swarm.ServiceSpec, headers map[string][]string) (types.ServiceCreateResponse, error)
 	ServiceInspectWithRaw(ctx context.Context, serviceID string) (swarm.Service, []byte, error)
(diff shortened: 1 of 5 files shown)


nishanttotla:
Ping @icecrime @vieux

vieux:
@nishanttotla to be consistant with the rest, should we create a

```go
type ServiceUpdateOptions {
        RegistryAuth  string // RegistryAuth is the base64 encoded credentials for the registry
}
``` 

in `types.go` and use that instead of the full headers map ?

nishanttotla:
@vieux I'm not sure if we want this to be the final way of handling auth tokens, so perhaps we should avoid creating a field which will just end up staying, and indicating that it is the right way to provide registry auth.

vieux:
ok then it's up to @docker/core-engine-api-maintainers

tiborvass:
It's just the client Go sdk, I'm fine with how it is now, if we add more options we can create the struct.

@nishanttotla needs rebase

nishanttotla:
@tiborvass done!

tiborvass:
LGTM

vieux:
LGTM

PR #295, opened 2016-06-30
Add directory as an option to checkpoint commands

boucher:
containerd provides support for providing the storage location for checkpoints, and this exposes that same support at the docker level. (This is useful when you want to create multiple containers from a single checkpoint).

Signed-off-by: Ross Boucher <rboucher@gmail.com>

### DIFF
--- a/client/checkpoint_delete.go
+++ b/client/checkpoint_delete.go
@@ -1,12 +1,20 @@
 package client
 
 import (
+	"net/url"
+
+	"github.com/docker/engine-api/types"
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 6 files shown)


cpuguy83:
Honestly, I think I'm 👎  on this.
Storage of checkpoints should be managed by Docker, not by a docker user.

boucher:
So, by default storage is managed by Docker. I feel like this fits in with the notion of batteries included but swappable. Not allowing you to specify the location of checkpoints is actually very limiting, and will basically force people to just go in and muck around anyway, with added expensive move operations.

crosbymichael:
Thanks for this PR.

PR #297, opened 2016-07-01
client/services: remove header arguments from ServiceCreate, ServiceU…

stevvooe:
In general, the Docker API client should avoid exposing HTTP details in
client, as it elides type safety and makes discovery of inputs to a
method more challenging. These have been removed and replaced with an
options type.

We may want to consider arbitrary header transmission through context,
but that is a bigger change.

Proper build tags have been added to experimental features that were
failing during non-tagged build.

@nishanttotla @tiborvass

### DIFF
--- a/client/service_create.go
+++ b/client/service_create.go
@@ -9,7 +9,15 @@ import (
 )
 
 // ServiceCreate creates a new Service.
-func (cli *Client) ServiceCreate(ctx context.Context, service swarm.ServiceSpec, headers map[string][]string) (types.ServiceCreateResponse, error) {
+func (cli *Client) ServiceCreate(ctx context.Context, service swarm.ServiceSpec, options types.ServiceCreateOptions) (types.ServiceCreateResponse, error) {
+	var headers map[string][]string
+
+	if options.EncodedRegistryAuth != "" {
+		headers = map[string][]string{
+			"X-Registry-Auth": []string{options.EncodedRegistryAuth},
+		}
+	}
+
 	var response types.ServiceCreateResponse
 	resp, err := cli.post(ctx, "/services/create", nil, service, headers)
 	if err != nil {
--- a/client/service_update.go
+++ b/client/service_update.go
@@ -4,14 +4,26 @@ import (
 	"net/url"
 	"strconv"
 
+	"github.com/docker/engine-api/types"
 	"github.com/docker/engine-api/types/swarm"
 	"golang.org/x/net/context"
 )
 
 // ServiceUpdate updates a Service.
-func (cli *Client) ServiceUpdate(ctx context.Context, serviceID string, version swarm.Version, service swarm.ServiceSpec, headers map[string][]string) error {
-	query := url.Values{}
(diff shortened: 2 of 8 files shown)


types/client.go
```diff
@@ -246,13 +246,33 @@ type NodeListOptions struct {
 	Filter filters.Args
 }
 
+type ServiceCreateOptions struct {
```
nishanttotla: If this struct isn't going to have anything other than `EncodedRegistryAuth` in the future, should we name it something else? `ServiceCreateOptions` sounds like it should contain more options.

vdemeester:
@stevvooe `golint` error :angel: 
```
types/client.go:249:6: exported type ServiceCreateOptions should have comment or be unexported
types/client.go:264:6: exported type ServiceUpdateOptions should have comment or be unexported
```

Other than that, looks good :wink:

types/client.go
vdemeester: It's more about being consistent with all other `*Options` struct.

dnephin: It's also very likely that it will gain other options over time.

nishanttotla: Okay fair enough, thanks!

nishanttotla:
LGTM, after CI is green.

client/service_create.go
```diff
@@ -9,7 +9,15 @@ import (
...
-func (cli *Client) ServiceCreate(ctx context.Context, service swarm.ServiceSpec, headers map[string][]string) (types.ServiceCreateResponse, error) {
+func (cli *Client) ServiceCreate(ctx context.Context, service swarm.ServiceSpec, options types.ServiceCreateOptions) (types.ServiceCreateResponse, error) {
+	var headers map[string][]string
+
+	if options.EncodedRegistryAuth != "" {
+		headers = map[string][]string{
+			"X-Registry-Auth": []string{options.EncodedRegistryAuth},
+		}
+	}
+
```
nishanttotla: This could just be `headers["X-Registry-Auth"] = []string{options.EncodedRegistryAuth}`, since headers has been declared above.

client/service_update.go
```diff
@@ -4,14 +4,26 @@ import (
...
+	var (
+		headers map[string][]string
+		query   = url.Values{}
+	)
+
+	if options.EncodedRegistryAuth != "" {
+		headers = map[string][]string{
+			"X-Registry-Auth": []string{options.EncodedRegistryAuth},
+		}
+	}
```
nishanttotla: Same here, `headers["X-Registry-Auth"] = []string{options.EncodedRegistryAuth}`

client/service_create.go
stevvooe: It hasn't been allocated above. We only do the allocation if the headers are required.

client/service_update.go
stevvooe: See above.

types/client.go
stevvooe: This looks funny now, but you'll thank me later.

client/service_create.go
nishanttotla: My bad, you're right.

nishanttotla:
LGTM

tiborvass:
LGTM

tiborvass:
Ping @vdemeester

vdemeester:
LGTM 🐸

PR #313, opened 2016-07-13
types: new seccomp format

runcom:
engine-api changes related to https://github.com/docker/docker/pull/24510

/cc @justincormack 

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

### DIFF
--- a/types/seccomp.go
+++ b/types/seccomp.go
@@ -2,12 +2,22 @@ package types
 
 // Seccomp represents the config for a seccomp profile for syscall restriction.
 type Seccomp struct {
-	DefaultAction Action     `json:"defaultAction"`
-	Architectures []Arch     `json:"architectures"`
-	Syscalls      []*Syscall `json:"syscalls"`
+	DefaultAction Action `json:"defaultAction"`
+	// Architectures is kept to maintain backward compatibility with the old
+	// seccomp profile.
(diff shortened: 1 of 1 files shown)


types/seccomp.go
```diff
@@ -2,12 +2,20 @@ package types
 
 // Seccomp represents the config for a seccomp profile for syscall restriction.
 type Seccomp struct {
-	DefaultAction Action     `json:"defaultAction"`
-	Architectures []Arch     `json:"architectures"`
-	Syscalls      []*Syscall `json:"syscalls"`
+	DefaultAction Action         `json:"defaultAction"`
+	Architectures []Arch         `json:"architectures,omitempty"`
+	ArchMap       []Architecture `json:"archMap,omitempty"`
```
vdemeester: This looks like a bit of duplication (or I don't really get it :sweat_smile:), but I guess this is for retro-compatibility right ? (if yes, a comment about that would be perfect :stuck_out_tongue_closed_eyes: )

vdemeester:
Quick question, otherwise LGTM 🐸

types/seccomp.go
justincormack: Yes, it is designed to be fully compatible.

justincormack:
LGTM

Issue #89, opened 2016-02-13
client.ImagePull needs a JSON message type


tamird:
Currently the only way to deserialize events from the `io.ReadCloser` returned from [`(*Client).ImagePull`](https://godoc.org/github.com/docker/engine-api/client#Client.ImagePull) is to use [`github.com/docker/docker/pkg/jsonmessage.JSONMessage`](https://godoc.org/github.com/docker/docker/pkg/jsonmessage#JSONMessage) which forces a dependency on the docker repo. Here's what code currently has to look like:

```go
	rc, err := cli.ImagePull(context.Background(), types.ImagePullOptions{
		ImageID: containerConfig.Image,
		Tag:     tag,
	}, nil)
	if err != nil {
		return err
	}
	defer rc.Close()
	dec := json.NewDecoder(rc)
	for {
		var message jsonmessage.JSONMessage
		if err := dec.Decode(&message); err != nil {
			if err == io.EOF {
				break
			}
			return err
		}
		log.Infof("ImagePull response: %s", message)
	}
```

Please introduce a new type in the engine-api repo (akin to [`github.com/docker/engine-api/types/events.Message`](https://godoc.org/github.com/docker/engine-api/types/events#Message)) that would allow deserializing these events.

vdemeester:
I guess we could bring `JSONMessage` to the engine-api, @calavera @MHBauer wdyt ?

tamird:
[`github.com/docker/engine-api/types/events.Message`](https://godoc.org/github.com/docker/engine-api/types/events#Message) contains a comment that indicates some deprecated fields are copied from `JSONMessage`. Is there an analogous custom message type planned for `ImagePull`?
```go
type Message struct {
    // Deprecated information from JSONMessage.
    // With data only in container events.
    ...
}
```

calavera:
I'd rather not bring JSONMessage here. There are a few packages in docker that depend on it, which would create a weird dependency.

Personally, I think JSONMessage is a bad generalization of a bunch of things we use in docker. I'd rather have something more clean. I'd rather have a clean implementation that's independent from Docker's package.

LK4D4:
I agree, JSONMessage is awful.

stevvooe:
This should be hidden behind the Go API. One should not be deserializing a raw `io.ReadCloser`. The signature should be something like:

```
ImagePull(ctx context.Context, name string, opts ImagePullOptions) (Progress, error)
```

Where `Progress` is some structure reporting the progress of the pull.

[...]

PR #315, opened 2016-07-13
Fixes #246 - moved event stream processing to engine api

### LINKED ISSUE #246: Add event management
[The main thing missing from this library that is stopping people from switching from samalba/dockerclient is event management.](https://github.com/samalba/dockerclient/issues/238)

@vdemeester has created [a separate library](https://github.com/vdemeester/docker-events) for parsing and handling events. Something like this should probably be in this library.

jhorwit2:
Signed-off-by: Joshua Horwitz <horwitzja@gmail.com>

Move the events logic to the engine-api while keeping backwards compatibility with the original `Events` method. Porting this logic to the engine-api makes it easier to write libraries utilizing the events code. I was having to use both docker core & engine-api while writing a library. Docker core has some dependencies that [aren't needed by external libraries](https://github.com/docker/docker/blob/master/api/client/system/events.go).

### DIFF
--- a/client/events.go
+++ b/client/events.go
@@ -1,20 +1,69 @@
 package client
 
 import (
-	"io"
+	"encoding/json"
 	"net/url"
 	"time"
 
 	"golang.org/x/net/context"
 
 	"github.com/docker/engine-api/types"
+	"github.com/docker/engine-api/types/events"
 	"github.com/docker/engine-api/types/filters"
 	timetypes "github.com/docker/engine-api/types/time"
 )
 
-// Events returns a stream of events in the daemon in a ReadCloser.
-// It's up to the caller to close the stream.
-func (cli *Client) Events(ctx context.Context, options types.EventsOptions) (io.ReadCloser, error) {
+// Events returns a stream of events in the daemon. It's up to the caller to close the stream
+// by cancelling the context. Once the stream has been completely read an io.EOF error will
+// be sent over the error channel. If an error is sent all processing will be stopped. It's up
+// to the caller to reopen the stream in the event of an error by reinvoking this method.
+func (cli *Client) Events(ctx context.Context, options types.EventsOptions) (<-chan events.Message, <-chan error) {
+
+	messages := make(chan events.Message)
+	errs := make(chan error, 1)
+
+	go func() {
+		defer close(errs)
+
+		query, err := buildEventsQueryParams(cli.version, options)
+		if err != nil {
+			errs <- err
+			return
+		}
+
+		resp, err := cli.get(ctx, "/events", query, nil)
+		if err != nil {
+			errs <- err
+			return
+		}
+		defer resp.body.Close()
+
+		decoder := json.NewDecoder(resp.body)
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				var event events.Message
+				if err := decoder.Decode(&event); err != nil {
+					errs <- err
+					return
+				}
+
+				select {
+				case messages <- event:
+				case <-ctx.Done():
+					return
+				}
+			}
+		}
+	}()
+
+	return messages, errs
+}
+
+func buildEventsQueryParams(cliVersion string, options types.EventsOptions) (url.Values, error) {
 	query := url.Values{}
 	ref := time.Now()
 
@@ -25,6 +74,7 @@ func (cli *Client) Events(ctx context.Context, options types.EventsOptions) (io.
 		}
 		query.Set("since", ts)
 	}
+
 	if options.Until != "" {
 		ts, err := timetypes.GetTimestamp(options.Until, ref)
 		if err != nil {
@@ -32,17 +82,14 @@ func (cli *Client) Events(ctx context.Context, options types.EventsOptions) (io.
 		}
 		query.Set("until", ts)
 	}
+
 	if options.Filters.Len() > 0 {
-		filterJSON, err := filters.ToParamWithVersion(cli.version, options.Filters)
+		filterJSON, err := filters.ToParamWithVersion(cliVersion, options.Filters)
 		if err != nil {
 			return nil, err
 		}
 		query.Set("filters", filterJSON)
 	}
 
-	serverResponse, err := cli.get(ctx, "/events", query, nil)
-	if err != nil {
-		return nil, err
-	}
-	return serverResponse.body, nil
+	return query, nil
 }
--- a/client/events_test.go
+++ b/client/events_test.go
@@ -2,7 +2,9 @@ package client
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"strings"
@@ -11,6 +13,7 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/docker/engine-api/types"
+	"github.com/docker/engine-api/types/events"
 	"github.com/docker/engine-api/types/filters"
 )
 
@@ -36,7 +39,8 @@ func TestEventsErrorInOptions(t *testing.T) {
 		client := &Client{
 			transport: newMockClient(nil, errorMock(http.StatusInternalServerError, "Server error")),
 		}
-		_, err := client.Events(context.Background(), e.options)
+		_, errs := client.Events(context.Background(), e.options)
+		err := <-errs
 		if err == nil || !strings.Contains(err.Error(), e.expectedError) {
 			t.Fatalf("expected a error %q, got %v", e.expectedError, err)
 		}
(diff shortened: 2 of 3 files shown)


vdemeester:
This is also related #246

jhorwit2:
@vdemeester I noticed that after starting to work on this PR. I noticed your lib takes into account types; do you think I should incorporate that into this PR? I initially was trying to keep it close to how core docker has it. 

The only thing I was thinking that should be different if we did add type/action handlers is to use a struct for keys over strings to make sure we don't have a collisions in the future. Another option is to just add a wildcard action instead of adding types. Then you can use the event options to set `{"type": {"container": true}}`, then `eh.Handle('*', cb)` or something.

stevvooe:
Consider some of the suggestions in https://github.com/docker/engine-api/issues/89#issuecomment-198570221 for an API. I'm not sure if a callback-based API is the correct approach if you have to listen for errors on a channel anyways. This pattern is employed [here](https://github.com/docker/swarmkit/blob/master/agent/exec/container/adapter.go#L141).

client/events.go
```diff
@@ -46,3 +47,48 @@ func (cli *Client) Events(ctx context.Context, options types.EventsOptions) (io.
...
+					// end of input stream reached
+					return
+				}
+
+				errChan <- err
+				return
+			case event := <-eventsq:
+				processer := handler.Get(event.Action)
+				if processer != nil {
+					processer(event)
```
jhorwit2: Keep going back and forth here on whether or not to do this in another goroutine. Thoughts? I like _not_ because then you keep to sequential processing which can be an important thing with events.

jhorwit2:
@stevvooe updated from your feedback 👍

client/events.go
stevvooe: That won't provide much. Serialization is cheap.

client/events.go
```diff
@@ -46,3 +47,48 @@ func (cli *Client) Events(ctx context.Context, options types.EventsOptions) (io.
 	}
 	return serverResponse.body, nil
 }
+
+// EventsWithHandler decodes the stream of events handing them off to the appropriate
+// handlers for processing. It's up to the caller to stop the processing by canceling
+// the context. The error channel will be closed once the end of the input stream has been
+// reached. If an error is received all processing will be stopped.
+func (cli *Client) EventsWithHandler(ctx context.Context, options types.EventsOptions, handler eventtypes.EventHandler) <-chan error {
```
stevvooe: I'd rather see something like this:

```
Events(ctx context.Context, options types.EventsOptions) (<-chan eventtypes.Message, <-chan error)
```

Then, we can have a wrapper built with a handler on top of this.

The issue with the handler approach is that it requires too much boilerplate to get something useful. A handler system can be built on the above, but it is hard to readapt the handler system to channel based system, if that is required instead.

client/events.go
```diff
@@ -46,3 +47,48 @@ func (cli *Client) Events(ctx context.Context, options types.EventsOptions) (io.
...
+	errChan := make(chan error)
```
stevvooe: This should be a buffered channel. Only ever send on it once.

jhorwit2: what value does making this `errChan := make(chan error, 1)` add? 

Currently it's only ever sent 1 error before the goroutine is returned from below.

stevvooe: The entire goroutine will block forever and leak if the caller does not listen on `errChan`.

jhorwit2: That makes sense. So it's a cheap way to program defensively if someone for some reason doesn't consume from the error chan. Thanks!

client/events.go
jhorwit2: You put `Events` as the name. Can I replace the legacy method with this new one? I figured we needed backwards compatibility so I went with another method name.

stevvooe: > Can I replace the legacy method with this new one? I figured we needed backwards compatibility so I went with another method name.

All the users of the `engine-api` vendor the dependency, so please feel free to change.

As it is, the method is very hard to use, so replacing the definition is fine. We'll stabilize these eventually, but we need to have an exported API worth stabilizing and we're not there yet.

jhorwit2:
@stevvooe i think this is what you had in mind 👍

stevvooe:
@jhorwit2 Love the new `Events` signature, but the handler needs more design work.

Do you mind splitting the handler parts into a second PR?

stevvooe:
Mostly, I'd like to get the Events signature merged... 👍 👍

jhorwit2:
Sounds good! Just pushed w/o the handler code @stevvooe.

stevvooe:
@jhorwit2 Thanks for the awesome contribution!

LGTM

Presumably, wait until have 1.12 GA until we merge this but it is up to engine-api maintainers.

cc @aaronlehmann @vdemeester

vdemeester:
@jhorwit2 Looks very good, thanks :heart_eyes_cat: 
LGTM 🐸

But let's wait the 1.12 release before merging this 👼

vdemeester:
@jhorwit2 would you like to make a PR in docker/docker with these changes to see how it behaves ? :angel:

jhorwit2:
@vdemeester yea sure! ill cherry pick the commit in and open a pr with vendor changes.

client/events.go
```diff
@@ -1,20 +1,70 @@
...
-// It's up to the caller to close the stream.
-func (cli *Client) Events(ctx context.Context, options types.EventsOptions) (io.ReadCloser, error) {
+// Events returns a stream of events in the daemon. It's up to the caller to close the stream
+// by cancelling the context. Once the stream has been completely read an io.EOF error will
+// be sent over the error channel. If an error is sent all processing will be stopped. It's up
+// to the caller to reopen the stream in the event of an error by reinvoking this method.
+func (cli *Client) Events(ctx context.Context, options types.EventsOptions) (<-chan events.Message, <-chan error) {
+
+	messages := make(chan events.Message, 1)
+	errs := make(chan error, 1)
```
jhorwit2: @stevvooe You mentioned in previous comment to make this a buffered channel in case the error wasn't consumed, but when you do this you force the consumer to check the event channel after an error has been received because both the event and error channel can have data. 

Ex)
```
for {
  evt := <- eventq:
     // Do something
  err := <- errq:
     return err
     // it's possible you just missed the last event
}
```

```
for {
  evt := <- eventq:
     // Do something
  err := <- errq:
     // boo error
     select {
          case evt := <- eventq:
               // make sure no left over events
          default:
               // guess no events
     }
     return err
}
```

That seems not great for someone using this method since it's not clear you need to do this w/o looking at the source code.

stevvooe: `messages` shouldn't be buffered, only `errq`.

jhorwit2: Derp... good catch.

stevvooe:
LGTM

PTAL @tonistiigi @aaronlehmann

client/events_test.go
```diff
@@ -98,29 +121,45 @@ func TestEvents(t *testing.T) {
...
 					actual := query.Get(key)
 					if actual != expected {
 						return nil, fmt.Errorf("%s not set in URL query properly. Expected '%s', got %s", key, expected, actual)
 					}
 				}
+
+				buffer := new(bytes.Buffer)
+
+				for _, e := range eventsCase.events {
+					b, _ := json.Marshal(e)
```
tamird: is it really worth it here to not check the error?

tamird: Also, it's probably more efficient to use a json.Encoder:
```
buffer := new(bytes.Buffer)
enc := json.NewEncoder(buffer)
for _, e := eventsCase.events {
  if err := enc.Encode(e); err != nil {
    // do something
  }
}
```

tamird: just realized this is the test! still, would be good to check the error.

tonistiigi:
Isn't it bit awkward to return 2 channels? Also when using select on these 2 channels it is easy to skip last messages because you could get the error first. Why not return a object instead with method for getting the events or error. Or channel returning union of these fields.

jhorwit2:
@tonistiigi because the message channel isn't buffered that won't happen

tamird:
Yeah, an object with a Next() method that returns `(events.Message, error)` might make more sense.

stevvooe:
@tamird That would disable the ability to select on a channel while consuming. You could always build a wrapper that does the `Next()` pattern.

tamird:
That's true.

client/events.go
```diff
@@ -1,20 +1,69 @@
...
+		}
+		defer resp.body.Close()
+
+		decoder := json.NewDecoder(resp.body)
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
```
aaronlehmann: This pattern makes sense, but I would end this `select` here to reduce unnecessary indentation.

aaronlehmann:
LGTM

jhowardmsft:
LGTM. Thanks!

jhorwit2:
So what's the plan with this since engine-api is now deprecated?

tonistiigi:
@jhorwit2 Just combine it with #25853 and copy this code to `github.com/docker/docker/client`. The use case of that package remains the same, we still expect other projects to import it if they want a docker client library.

vdemeester:
anyway let's close this now as it is all in docker/docker#25853

PR #316, opened 2016-07-14
refactor function getDockerOS() in filegithub.com/docker/engine-api/c…

yuanxiaoaini:
…lient/image_build.go
- What I did
refactor function “getDockerOS()” in file github.com/docker/engine-api/client/image_build.go
- How I did it
1. remove the local variable -- ”var osType string“

    refactor the statement "if len(matches) > 0 {osType = matches[1]}", because of when len(matches) == 1 , it will throw exception (index out of range)

- How to verify it
local unit test

- Description for the changelog

refactor the statement "if len(matches) > 0 {osType = matches[1]}", because of when len(matches) == 1 , it will throw exception (index out of range). More Robust !!!
Signed-off-by: yuanxiao <yuan.xiao5@zte.com.cn>

### DIFF
--- a/client/image_build.go
+++ b/client/image_build.go
@@ -110,10 +110,9 @@ func imageBuildOptionsToQuery(options types.ImageBuildOptions) (url.Values, erro
 }
 
 func getDockerOS(serverHeader string) string {
-	var osType string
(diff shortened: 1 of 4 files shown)


yuanxiaoaini:
sorry something wrong with my branch it do not work with your guides

PR #317, opened 2016-07-14
refactor functions in folder github.com/docker/engine-api/client/

SinghamXiao:
Signed-off-by: yuanxiao <yuan.xiao5@zte.com.cn>

### DIFF
--- a/client/image_build.go
+++ b/client/image_build.go
@@ -110,10 +110,9 @@ func imageBuildOptionsToQuery(options types.ImageBuildOptions) (url.Values, erro
 }
 
 func getDockerOS(serverHeader string) string {
-	var osType string
 	matches := headerRegexp.FindStringSubmatch(serverHeader)
-	if len(matches) > 0 {
-		osType = matches[1]
+	if len(matches) < 2 {
+		return ""
 	}
-	return osType
+	return matches[1]
 }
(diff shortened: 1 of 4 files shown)


vdemeester:
Hi @yuanxiaoaini, thanks for the contribution but could you provide more context on the changes (in the description of the PR on the changes).

Concerning `getDockerOS`, could you add a test case for `ImageBuild` that shows why the change is needed.

I'm really not a fan of the other changes. Named return are a pain to deal with and making return explicity is easier to read for a human I feel 👼

yuanxiaoaini:
Thank you for your review.
Actually, Nobody can construct the test that " len(matches) == 1 ",  because of the " matches " returns from " matches := headerRegexp.FindStringSubmatch(serverHeader) " only have two kind of values ( nil or slice which length is 2 ). 
However, I'm really do not think the statement of " if len(matches) > 0 {osType = matches[1]} " is a good programming habits. It is hard for me to understand that when " len(matches) > 0 ", " matches[1] " is used.

cpuguy83:
-1 on the named returns.

vdemeester:
ping @SinghamXiao those changes now needs to be done in docker/docker directly (the `client`'s package)

crosbymichael:
Thanks for this PR.

PR #318, opened 2016-07-14
Remove SwarmInspect method.

vdemeester:
SwarmInspect information are redundant with Info. The API won't serve
`GET /swarm` anymore so removing the client method too. 👼

Related PR https://github.com/docker/docker/pull/24492

/cc @thaJeztah @icecrime @tiborvass @aluzzardi @stevvooe 

🐸


thaJeztah:
LGTM

icecrime:
LGTM

types/swarm/swarm.go
```diff
@@ -120,6 +120,8 @@ type Info struct {
 	Nodes          int
 	Managers       int
 	CACertHash     string
+
+	Cluster Swarm
```
stevvooe:
```
// Cluster contains data about the swarm to which this node has joined.
````

stevvooe:
Are we making a mistake by creating a compound data structures? `Info` is really about the node's local configuration, whereas cluster is about the configuration of the cluster.

The main issue here is that we end up with an imbalance between the object that is modified and how to read the results of that modification.

vdemeester:
> The main issue here is that we end up with an imbalance between the object that is modified and how to read the results of that modification.

Yeah, I do agree on that one, it feels a little weird to update swarm properties  via `/swarm/*` and getting swarm properties through `/info`.. 

Maybe we should keep swarm inspect and endpoint and *just* make `docker info` call the endpoint.

stevvooe:
@tiborvass This is not back and forth. This is CLI decisions being used to make thoughtless changes to the API. We have a deep design problem that our API and CLI are over-coupled. I'm fine with the CLI decision. It is a good one. I just don't understand why we think that means we have to delete this endpoint on the API.

Go ahead and merge this.

aluzzardi:
Hey, side note - I think we need to be able to inspect in order to update (which we use for swarm config, such as join secret etc)

PR #319, opened 2016-07-14
Add list of failed task on service inspect

vieux:
ping @aluzzardi @icecrime @stevvooe @tiborvass @dnephin

### DIFF
--- a/client/interface.go
+++ b/client/interface.go
@@ -101,7 +101,7 @@ type NodeAPIClient interface {
 // ServiceAPIClient defines API client methods for the services
 type ServiceAPIClient interface {
 	ServiceCreate(ctx context.Context, service swarm.ServiceSpec, options types.ServiceCreateOptions) (types.ServiceCreateResponse, error)
-	ServiceInspectWithRaw(ctx context.Context, serviceID string) (swarm.Service, []byte, error)
+	ServiceInspectWithRaw(ctx context.Context, serviceID string) (swarm.InspectService, []byte, error)
 	ServiceList(ctx context.Context, options types.ServiceListOptions) ([]swarm.Service, error)
 	ServiceRemove(ctx context.Context, serviceID string) error
 	ServiceUpdate(ctx context.Context, serviceID string, version swarm.Version, service swarm.ServiceSpec, options types.ServiceUpdateOptions) error
--- a/client/service_inspect.go
+++ b/client/service_inspect.go
@@ -11,22 +11,22 @@ import (
 )
 
 // ServiceInspectWithRaw returns the service information and the raw data.
-func (cli *Client) ServiceInspectWithRaw(ctx context.Context, serviceID string) (swarm.Service, []byte, error) {
+func (cli *Client) ServiceInspectWithRaw(ctx context.Context, serviceID string) (swarm.InspectService, []byte, error) {
 	serverResp, err := cli.get(ctx, "/services/"+serviceID, nil, nil)
 	if err != nil {
 		if serverResp.statusCode == http.StatusNotFound {
-			return swarm.Service{}, nil, serviceNotFoundError{serviceID}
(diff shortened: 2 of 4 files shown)


vdemeester:
I think related `docker/docker` issue is https://github.com/docker/docker/issues/23706 (and https://github.com/docker/docker/issues/23811) 👼

LGTM 🐳

stevvooe:
👎 If we are going to depart from the swarmkit model, we need to have different type. There is already a task endpoint to get failed tasks. This is the wrong approach.

vieux:
@aluzzardi @icecrime @stevvooe @tiborvass @dnephin how about this ?

stevvooe:
This data is already available through the tasks API. Why have two ways to get it?

It also reenforces the model that services have failures, when it is tasks that have failures.

I really think we need to have `docker service ps` cover this use case.

dnephin:
LGTM

I understand from an object modelling perspective it is not the nicest solution, but for UX `service inspect` is the most natural place to find these errors. 

Either we need to break the convention of `inspect` returning a single API object, or we need this change. It sounds like others prefer this option.

stevvooe:
> I understand from an object modelling perspective it is not the nicest solution, but for UX service inspect is the most natural place to find these errors.

Other than the simple fact that services don't have errors and here we are, teaching users that services have errors.

Let's have a command that displays errors.

thaJeztah:
@stevvooe But a service can have an error in its definition, causing the tasks to fail.
What would be the workflow (from a user-perspective) if I create or update a service, and I see that my replicas stick at 0? What steps should I take to find the error?

(not trolling, just curious how that would work if we take the "correct" model)

stevvooe:
@thaJeztah A service of that nature has a definition such that tasks created for that service result in errors. The issue here is that a misconfiguration may only be temporary. One example is that an image may not exist or be available at service creation time, but may be available later.

We say that such a service cannot "converge".

The appropriate troubleshooting methodology would be as follows:

1. Observe through UI that the service is non-converging (we could report using heuristics at the client or in orchestrator).
2. Investigate cause of task failures.

In `swarmkit`, this is very natural, since the concept of `inspect` is more of a status. Unfortunately, `docker inspect` isn't assembled with multiple API calls and default output isn't very usable, since it is JSON.

We can see the design mistake when we look at `swarmctl service inspect`:

```
$ bin/swarmctl service inspect echo
ID                : 3nb5zdmfrdv5g61zyfttxp8nh
Name              : echo
Replicas          : 1/1
Template
 Container
  Image           : alpine
  Command         : "sleep 1000"

Task ID                      Service    Slot    Image     Desired State    Last State                Node
-------                      -------    ----    -----     -------------    ----------                ----
bqa6jhjodm9rsv6onyhlj2rwd    echo       1       alpine    RUNNING          RUNNING 14 seconds ago    moby
```

If we update this to the wrong image:

```
$ bin/swarmctl service update --image wrongimage echo
$ bin/swarmctl service inspect echo
ID                : 3nb5zdmfrdv5g61zyfttxp8nh
Name              : echo
Replicas          : 0/1
Template
 Container
  Image           : wrongimage
  Command         : "sleep 1000"

Task ID                      Service    Slot    Image         Desired State    Last State                Node
-------                      -------    ----    -----         -------------    ----------                ----
abxcfkjrp6x9jno1hpyq7d3zm    echo       1       wrongimage    ACCEPTED         ACCEPTED 3 seconds ago    moby
```

We'll actually start seeing this flap between rejected and pre-starting states. One can drill down by inspecting tasks and see the error history, but we could show that information here, as well.

The solution here is a user-friendly command that is an assembly of API calls. I propose `docker service status`. This will let us experiment with the proper UX and not introduce hasty, incorrect changes into the API.

icecrime:
LGTM.

vieux:
@aluzzardi found another way to display errors in a meaningful way without changing the API. I'm closing this one.

PR #320, opened 2016-07-15
Make darwin default docker host the same as unix

xetorthio:
The reason for this is that docker already default for unix docker in darwin too.
Also now that we have Docker for Mac, we can talk to the daemon using a unix socket.
So it seems like it makes a lot more sense that unix docker is the default.

### DIFF
--- a/client/client_darwin.go
+++ b/client/client_darwin.go
@@ -1,4 +1,0 @@
-package client
-
-// DefaultDockerHost defines os specific default if DOCKER_HOST is unset
-const DefaultDockerHost = "tcp://127.0.0.1:2375"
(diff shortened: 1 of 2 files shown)


marcosnils:
ref: https://github.com/docker/docker/blob/master/opts/hosts_unix.go

tamird:
Ping, this is also affecting us. Also, did this ever work before? AFAICT the old docker-on-mac-via-docker-machine solution relied on `DOCKER_HOST` being set, so this was never correct.

vdemeester:
LGTM 🐸
/cc @thaJeztah @justincormack

justincormack:
LGTM

Also unclear in what environment this was ever useful...

thaJeztah:
LGTM, thanks!

PR #321, opened 2016-07-15
Add `--force` in `docker volume rm` to fix out-of-band volume driver deletion

yongtang:
This fix is related to pull request in docker:

https://github.com/docker/docker/pull/23436

And this fix tries to address the issue raised in

https://github.com/docker/docker/issues/23367

where an out-of-band volume driver deletion leaves some data in docker. This prevent the reuse of deleted volume names (by out-of-band volume driver like flocker).

This fix adds a `--force` field in `docker volume rm` to forcefully purge the data of the volume that has already been deleted.


thaJeztah:
Thanks @yongtang 

For reviewers: we discussed this improvement during a maintainers meeting / review session, and didn't think it was required for 1.12, so don't merge yet (not before 1.12 GA)

/cc @cpuguy83

vdemeester:
ping @cpuguy83

cpuguy83:
LGTM

PR #326, opened 2016-07-20
Support advertise address and node address

aaronlehmann:
This makes the engine-api changes to support
https://github.com/docker/docker/pull/24237.

AdvertiseAddr is added to InitRequest and JoinRequest so a node can
specify a different address to advertise than the one it binds to. This
allows scenarios with NAT to be supported, and also allows binding to a
wildcard address.

NodeAddr in Info allows Docker to expose a node's address.


LK4D4:
LGTM

thaJeztah:
LGTM

PR #333, opened 2016-07-22
Add failure action for rolling updates

aaronlehmann:
This is the engine-api portion of docker/docker#24943.

cc @stevvooe @aluzzardi @cpuguy83 @tiborvass

### DIFF
--- a/types/swarm/service.go
+++ b/types/swarm/service.go
@@ -6,8 +6,9 @@ import "time"
 type Service struct {
 	ID string
 	Meta
-	Spec     ServiceSpec `json:",omitempty"`
-	Endpoint Endpoint    `json:",omitempty"`
+	Spec         ServiceSpec  `json:",omitempty"`
+	Endpoint     Endpoint     `json:",omitempty"`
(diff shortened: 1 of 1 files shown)


types/swarm/service.go
```diff
@@ -29,6 +30,26 @@ type ServiceMode struct {
 	Global     *GlobalService     `json:",omitempty"`
 }
 
+// UpdateState is the state of a service update.
+type UpdateState string
```
tiborvass: Why have a `UpdateState` type and not a `FailureAction` type?

aaronlehmann: I tried having a `FailureAction` type and it turned out to be cumbersome. The client couldn't handle it like a normal string because of the special type. But I can add it back if you think it should have its own type (or remove `UpdateState` - whatever makes more sense).

tiborvass: Not at all, was just curious

tiborvass:
LGTM

crosbymichael:
LGTM

PR #338, opened 2016-07-26
Rename the info LiveRestore field to LiveRestoreEnabled and document it

mlaventure:
Signed-off-by: Kenfe-Mickael Laventure <mickael.laventure@gmail.com>

--

ping @stevvooe

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -256,7 +256,10 @@ type Info struct {
 	Runtimes           map[string]Runtime
 	DefaultRuntime     string
 	Swarm              swarm.Info
-	LiveRestore        bool
+	// LiveRestoreEnabled determines whether containers should be kept
+	// running when the daemon is shutdown or upon daemon start if
+	// running containers are detected
+	LiveRestoreEnabled bool
 }
 
 // PluginsInfo is a temp struct holding Plugins name


types/types.go
```diff
@@ -256,7 +256,10 @@ type Info struct {
 	Runtimes           map[string]Runtime
 	DefaultRuntime     string
 	Swarm              swarm.Info
-	LiveRestore        bool
+	// Determine whether containers should be kept running when the
```
stevvooe: Please follow the [Go style guidelines](https://golang.org/doc/effective_go.html#commentary) when documenting fields.

mlaventure: Sure, but I can't see what I'm doing against the guideline there.

stevvooe: "The first sentence should be a one-sentence summary that starts with the name being declared."

This should be:

```
// LiveRestore determines ...
```

mlaventure: That's only for routine I think. From the page:
```
type T struct {
    name    string // name of the object
    value   int    // its value
}
```

But I don't mind updating it.

stevvooe: No. The entire standard library and all the code in docker follows (or should) this standard.

See https://github.com/golang/go/wiki/CodeReviewComments#doc-comments

mlaventure: I meant: the go official examples themselves don't enforce this for struct fields. The example I put in my previous comment, is straight from the go guideline you've linked. 

And yes, this is not enforced for struct fields in our repos or by golint for that matter (just look 10 lines down).

Anyhow, I'll change it.

stevvooe: https://golang.org/pkg/runtime/#MemProfileRecord
https://golang.org/pkg/crypto/tls/#Config
https://golang.org/pkg/image/#Alpha16
https://godoc.org/github.com/docker/distribution#Descriptor
https://golang.org/pkg/text/scanner/#Scanner

Document the fields of a struct so we don't have to waste time figuring out what they do.

mlaventure: :joy: I wasn't arguing the comment to the field, it makes perfect sense.

Just the fact that the go standard doesn't require comment for a field to start with the field name.

stevvooe: > Just the fact that the go standard doesn't require comment for a field to start with the field name.

Yea. I waste a lot of time asking people to document fields.

stevvooe:
LGTM

vdemeester:
LGTM 🐭

PR #339, opened 2016-07-27
Remove join-tokens from info

thaJeztah:
join-tokens are not needed for the endpoint

relates to https://github.com/docker/docker/pull/25134

### DIFF
--- a/types/swarm/swarm.go
+++ b/types/swarm/swarm.go
@@ -10,6 +10,14 @@ type Swarm struct {
 	JoinTokens JoinTokens
 }
 
+// ClusterInfo represents info about a the cluster for outputing in "info"
+// it contains the same information as "Swarm", but without the JoinTokens
+type ClusterInfo struct {
+	ID string
+	Meta
+	Spec Spec
+}
+
 // JoinTokens contains the tokens workers and managers need to join the swarm.
 type JoinTokens struct {
 	Worker  string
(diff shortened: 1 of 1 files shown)


thaJeztah:
golint didn't like my naming, so changing to `ClusterInfo`

tiborvass:
LGTM

aaronlehmann:
How about this to avoid duplication:

// ClusterInfo represents info about a the cluster for outputing in "info"
// if contains the same information as "Swarm", but without the JoinTokens
type ClusterInfo struct {
        ID string
        Meta
        Spec Spec
}

// Swarm represents a swarm.
type Swarm struct {
        ClusterInfo
        JoinTokens JoinTokens
}

thaJeztah:
pushing change in a minute

vieux:
@thaJeztah CI fail

thaJeztah:
changed back to my original approach; @aaronlehmann's suggestion requires more changes

vieux:
Do we need this for 1.12 ? it not, maybe we should take the time and implement @aaronlehmann's suggestion ?

thaJeztah:
@vieux we can keep it for 1.12.1; there's a danger that people will start to rely on this, and removing it after 1.12 may become more problematic then, but possibly by just not documenting the field :blush:

thaJeztah:
Alright discussed with @tiborvass, let's move this to 1.13

aaronlehmann:
Carried in #340.

PR #343, opened 2016-07-29
Add isolation to info struct

jhowardmsft:
Signed-off-by: John Howard <jhoward@microsoft.com>

Apart from being useful information to expose through `INFO` on the engine, this is necessary for the bring up of CI on Hyper-V containers on Windows. Some tests need to behave differently depending on the isolation mode of the daemon being tested (for example, in a Hyper-V container, you can't find and kill a container process from the host as it's running inside a utility VM).

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -260,6 +260,7 @@ type Info struct {
 	// running when the daemon is shutdown or upon daemon start if
 	// running containers are detected
 	LiveRestoreEnabled bool
+	Isolation          container.Isolation
 }
 
 // PluginsInfo is a temp struct holding Plugins name


thaJeztah:
LGTM

cpuguy83:
Being that windows and Linux have a similar concept here (Runtime on Linux), it may be nice to be able to use the same field for both... e.g. `Default<ThingThatRunsYourContainer>=(runc|hyper-v|...)`

thaJeztah:
> it may be nice to be able to use the same field for both

Didn't think of that (or forgot), and I like that idea; any objections against doing that @jhowardmsft ?

jhowardmsft:
Hmmm. These are actually slightly different. In theory, you can have different isolation modes on linux and still have the same runtime. And of course, we may introduce runc in the future on Windows which supports multiple isolations. So I really don't believe these should be cross-pollinated into the same field.

jhowardmsft:
ping @cpuguy83

jhowardmsft:
Please can we move this forward. We have no CI coverage on Hyper-V containers which is kinda important... :innocent:

jhowardmsft:
@cpuguy83 Can you please see my response to your comment above so that we can hopefully finally move this forward. Thanks.

cpuguy83:
@jhowardmsft Honestly, I'm -1 on the change. I don't think it's different enough and it's adding even greater divergence in the API between platforms when it probably doesn't need to.

jhowardmsft:
So how do you suggest doing this if/when there are multiple run times on Windows? Isolation is already in the API. This is purely extending that to info so really is not the concern you are raising...

jhowardmsft:
@thaJeztah Can you curate this then please and herd through the Thursday meetings/maintainers. This is blocking bringing up CI of Hyper-V Containers. As I mention above, it already in the API, quite separate (rightly) from runtime. eg https://github.com/docker/docker/blob/master/runconfig/opts/parse.go#L97-L104, and in places where runtime isn't applicable eg https://github.com/docker/docker/blob/master/api/client/image/build.go#L52. I completely disagree that the runtime field should be re-purposed or overloaded as it will just make a situation akin to unravelling binds/mounts in the future.  The platforms are already significantly diverged - round pegs and square holes will never meet, and this is just an extension of an existing divergence.

jhowardmsft:
Closing. Engine-api is now in docker/docker

PR #344, opened 2016-07-30
Add legacy option to network create

aboch:


cpuguy83:
What does this do exactly? Do you have a corresponding change on docker engine?

aboch:
Yes, it's https://github.com/docker/docker/pull/25263. I should have added it to the description.

Regarding the support for docker run on swarm networks, we need a way to constrain this to specific networks. it was thought of adding a new network field to mark the swarm networks on which we support running containers in the legacy way.

stevvooe:
Please don't do this. `Legacy` doesn't mean anything. We are just in the process of building a new legacy.

cpuguy83:
Agreed with @stevvooe, I would hope there would be a way to do this without adding a new field to the API.

Going to close this because I think this is just really bad design.

PR #351, opened 2016-08-04
replace .Active by .Enabled

vieux:
ping @tiborvass @anusha-ragunathan 

related to  https://github.com/docker/docker/pull/25424

### DIFF
--- a/types/plugin.go
+++ b/types/plugin.go
@@ -26,10 +26,11 @@ type PluginConfig struct {
 
 // Plugin represents a Docker plugin for the remote API
 type Plugin struct {
-	ID       string `json:"Id,omitempty"`
-	Name     string
-	Tag      string
-	Active   bool
+	ID   string `json:"Id,omitempty"`
+	Name string
+	Tag  string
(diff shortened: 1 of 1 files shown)


tiborvass:
*Cross posting from https://github.com/docker/docker/pull/25424#issuecomment-237715136:*
I wonder if it should be State: Enabled instead.

anusha-ragunathan:
LGTM

types/plugin.go
```diff
@@ -29,7 +29,7 @@ type Plugin struct {
 	ID       string `json:"Id,omitempty"`
 	Name     string
 	Tag      string
-	Active   bool
+	Enabled  bool
```
stevvooe: I know we don't do this in this project, but could we _please_ document our fields? What does enabled actually mean?

stevvooe:
LGTM after documenting the field

I am not convinced a state machine will help here, without developing a proper model around the state machine.

vieux:
PTAL @stevvooe

tiborvass:
LGTM

stevvooe:
LGTM

PR #356, opened 2016-08-09
client: pass `serverResponse` around by value

tamird:
Removes a bunch of impossible and useless nil checks.

Signed-off-by: Tamir Duberstein <tamird@gmail.com>

### DIFF
--- a/client/container_create.go
+++ b/client/container_create.go
@@ -34,7 +34,7 @@ func (cli *Client) ContainerCreate(ctx context.Context, config *container.Config
 
 	serverResp, err := cli.post(ctx, "/containers/create", query, body, nil)
 	if err != nil {
-		if serverResp != nil && serverResp.statusCode == 404 && strings.Contains(err.Error(), "No such image") {
+		if serverResp.statusCode == 404 && strings.Contains(err.Error(), "No such image") {
 			return response, imageNotFoundError{config.Image}
 		}
 		return response, err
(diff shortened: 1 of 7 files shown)


tamird:
@thaJeztah @calavera can you guys take a look? You appear to be the main contributors to `client/request.go`.

cpuguy83:
I like it, LGTM.

LK4D4:
@tamird It's sorta against the "rules"(like you're supposed to use pointer if your struct contains pointer). I think that it's impossible to have nil serverResp because it's instantiated at the top of `sendClientRequest`, so you can just remove nil-checks.

tamird:
Why are you supposed to use a pointer if the struct contains a pointer?

LK4D4:
@tamird It's code style and it's easier for GC to track such pointers I think.

LK4D4:
@tamird I thought that escape analysis will put them on heap anyway if they contain pointers, maybe I'm wrong.
I still think that it's better to use smaller diff with same effect, because serverResp never can be nil.

tamird:
@LK4D4 escape analysis will put the thing pointed *to* on the heap, not the thing containing the pointer.

No, it is better to have something which is more type safe (which this is, because it doesn't use a nullable pointer at all).

LK4D4:
@tamird ok, you're right
LGTM

PR #357, opened 2016-08-10
cancellable: use sync.Once to prevent double channel close

LK4D4:
ping @tonistiigi

### DIFF
--- a/client/transport/cancellable/cancellable.go
+++ b/client/transport/cancellable/cancellable.go
@@ -8,6 +8,7 @@ package cancellable
 import (
 	"io"
 	"net/http"
+	"sync"
 
 	"github.com/docker/engine-api/client/transport"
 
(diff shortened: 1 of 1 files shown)


tonistiigi:
@LK4D4 Would it make sense to just remove `notifyingReader` if it's not used for anything?

LK4D4:
@tonistiigi I noticed that that `Do` uses it. It waits on that channel there.

client/transport/cancellable/cancellable.go
```diff
@@ -91,23 +92,24 @@ func Do(ctx context.Context, client transport.Sender, req *http.Request) (*http.
...
+		r.notifyOnce.Do(func() {
+			close(r.notify)
+		})
 	}
 	return n, err
 }
 
 func (r *notifyingReader) Close() error {
 	err := r.ReadCloser.Close()
-	if r.notify != nil {
```
mlaventure: Shouldn't `r.notify` should still be check for `nil`? There's no warranty that the field was assigned to, is there?

tonistiigi: It's not part of a public api.

LK4D4: @mlaventure yes, @tonistiigi is right. Even more, it's used only in this file.

vdemeester:
LGTM :cat:

PR #361, opened 2016-08-12
Use dialer from client to set up hijacked connection

challengee:
After some digging with @devwout, we discovered that `postHijacked` just uses `net.Dial` instead of the one from the underlaying transport. We're having problems with this because we tunnel the docker socket over SSH.

The `tlsDialWithDialer` now used a `Dialer` interface instead of `net.Dialer`, which means we don't get all of the timeout options. However these were not used (`new(net.Dialer)`), so I dumped them.

There was no test for `TestContainerExecAttach`, I also didn't add one.

This probably needs some additional work though.

Thanks!

### DIFF
--- a/client/transport/client.go
+++ b/client/transport/client.go
@@ -2,8 +2,15 @@ package transport
 
 import (
 	"crypto/tls"
+	"net"
 	"net/http"
 )
+
+// Dialer is an interface thet clients must implement
+// to be able to set up hijacked connections.
+type Dialer interface {
+	Dial(string, string) (net.Conn, error)
+}
 
 // Sender is an interface that clients must implement
 // to be able to send requests to a remote connection.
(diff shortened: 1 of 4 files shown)


client/transport/client.go
```diff
@@ -14,6 +21,7 @@ type Sender interface {
 
 // Client is an interface that abstracts all remote connections.
 type Client interface {
+	Dialer
```
stevvooe: `Dialer` doesn't really belong at this level.

stevvooe: Actual dialers are configured by https://github.com/docker/go-connections/blob/master/sockets/sockets.go#L22.

There doesn't seem to be a good way to get to these without changing that project.

I think it is okay if this interface has a `Dialer() Dialer` method. Seems odd, but preserves the taxonomy here.

crosbymichael:
Thanks for this PR.

PR #362, opened 2016-08-12
Correct order of error checking on connect failure

cpuguy83:
1730fe8fb0452def2eafed4142bbd9139bda35db changed error checking to use
the `Temporary()` function from `net.Error`. Unfortunately having this
check so high up it is catching more than we really want it to and
causing the custom error checks below it to not be caught.

This moves the net.Error check to last since it is the most generic
check.

Signed-off-by: Brian Goff <cpuguy83@gmail.com>

### DIFF
--- a/client/request.go
+++ b/client/request.go
@@ -106,8 +106,12 @@ func (cli *Client) sendClientRequest(ctx context.Context, method, path string, q
 
 	resp, err := cancellable.Do(ctx, cli.transport, req)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cpuguy83:
ping @thaJeztah @vdemeester

cpuguy83:
btw, this fixes some tests in docker CI.

vdemeester:
ah 😅
LGTM 🐵

LK4D4:
LGTM

PR #366, opened 2016-08-15
Fixes more issues swallowing connection errors

cpuguy83:
Really need to make sure we are only returning `ErrConnectionFailed`
when it's really a failed connection, not just something that reports as
a non-temporary error.

This fixes more tests on docker CI, see https://jenkins.dockerproject.org/job/Docker-PRs/31060/console for reference


vdemeester:
LGTM 🐵

LK4D4:
LGTM

PR #382, opened 2016-09-02
Fix CI for go1.7

tonistiigi:
govet detected a case where we copy tls.Config although
the var contains mutexes. This is fixed in go1.8 with a
stdlib Clone method. Temporarily add a local copy of that 
function until we can move to go1.8 

Signed-off-by: Tonis Tiigi <tonistiigi@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,6 @@
   notifications:
     email: false
   go:
-    - 1.6
+    - 1.7
   install: make deps
   script: make validate && make test
--- a/client/hijack.go
+++ b/client/hijack.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/docker/engine-api/client/transport"
 	"github.com/docker/engine-api/types"
 	"github.com/docker/go-connections/sockets"
 	"golang.org/x/net/context"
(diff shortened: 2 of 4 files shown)


tonistiigi:
@tiborvass @aaronlehmann

tiborvass:
So this will not work for Go1.6 and under, I guess it's fine. LGTM

aaronlehmann:
Is this really how `Clone` is implemented upstream? I know we don't care for our purposes, but I would expect a function called `Clone` to make a deep copy of slices and maps inside the struct. Otherwise changes to the slice or map in the original object will affect the clone, or vice versa.

tonistiigi:
@tiborvass I can make a go1.6 function as well if you think it's important. We can't reuse the same function though because new keys were added.

@aaronlehmann https://github.com/golang/go/commit/d24f446a90ea94b87591bf16228d7d871fec3d92 https://github.com/golang/go/commit/db5802104797cadcb4f44c5198a0fc39e13f9bc3

aaronlehmann:
Fair enough: `Clone returns a shallow clone of c.`

aaronlehmann:
LGTM

